PR #1, opened 2020-02-07
DM-23355: Initial library implementation

jonathansick:
This sets up the basic packaging structure for Safir and curates functionality from existing Roundtable apps (https://github.com/lsst-sqre/sqrbot-jr and https://github.com/lsst-sqre/templatebot).

- [x] PEP 518 pyproject.toml with most configuration in setup.cfg
- [x] setuptools_scm for versioning.
- [x] Tox-based testing
- [x] Black and isort, run through pre-commit
- [x] GitHub Actions-based CI
- [x] Documentation infrastructure
- [x] Logging helpers (logging configuration and middleware to bind a structlog logger with request context to a handler)
- [x] Helpers to create an aiohttp session for the app
- [x] Helpers for configuring metadata
- ~~Helpers to create a Kafka producer and consumer~~ → will move to a subsequent PR
- [x] Safir developer documentation
- [x] A Getting-started tutorial for using the `roundtable_aiohttp_bot` template

The preview documentation is at: https://safir.lsst.io/v/DM-23355/

Some of the documentation that's needed in the short term, but isn't there yet are:

- Guide for application configuration (to document Safir's conventions)
- Guide for application metadata (though there are docstrings)
- Guide for adding routes


pyproject.toml
```diff
@@ -0,0 +1,90 @@
...
+  | dist
+)/
+'''
+# Use single-quoted strings so TOML treats the string like a Python r-string
+# Multi-line strings are implicitly treated by black as regular expressions
+
+[tool.isort]
+multi_line_output = 3
+known_first_party = "safir"
+known_third_party = ["aiohttp", "safir", "setuptools", "structlog"]
```
rra: safir is listed as both first-party and third-party. Is that intended?

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+project_urls =
+    Change log = https://safir.lsst.io/changelog.html
+    Source code = https://github.com/lsst-sqre/safir
+    Issue tracker = https://github.com/lsst-sqre/safir/issues
+classifiers =
+    Development Status :: 4 - Beta
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.7
```
rra: You also test under Python 3.8, so maybe list that as well?

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    Operating System :: POSIX
+keywords =
+    lsst
+
+[options]
+zip_safe = False
+include_package_data = True
+package_dir =
+    = src
+packages=find:
```
rra: Nit: whitespace around `=`?

src/safir/http.py
```diff
@@ -0,0 +1,48 @@
...
+
+from typing import TYPE_CHECKING
+
+from aiohttp import ClientSession
+
+if TYPE_CHECKING:
+    from aiohttp.web import Application
+
+
+async def init_http_session(app: "Application"):
```
rra: Add `from __future__ import annotations` to the top of the file and then you can stop quoting the types that are only imported under `if TYPE_CHECKING`.

src/safir/http.py
```diff
@@ -0,0 +1,48 @@
...
+        The aiohttp.web-based application.
+
+    Notes
+    -----
+    Use this function as a `cleanup context
+    <https://aiohttp.readthedocs.io/en/stable/web_reference.html#aiohttp.web.Application.cleanup_ctx>`__:
+
+    .. code-block:: python
+
+       python.cleanup_ctx.append(init_http_session)
```
rra: Should this be `app.cleanup_ctx.append`?

src/safir/logging.py
```diff
@@ -0,0 +1,172 @@
...
+Examples
+--------
+Usage:
+
+>>> logger = response_logger.get()
+>>> logger.info(key='value')
+"""
+
+
+def get_response_logger() -> "BoundLoggerLazyProxy":
```
rra: Same note here on `from __future__ import annotations`.

src/safir/logging.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Examples
+    --------
+    Usage:
+
+    .. code-block:: python
+
+       from safir.logging import get_response_logger
+       logger = get_response_logger()
+       logger.info('Some message', somekey='somevalue')
```
rra: Nit: You're using single quotes in the Python examples in the documentation, but black will fix all quoting to use double quotes. Might be worth using double quotes in the documentation examples as well for consistency.

src/safir/middleware.py
```diff
@@ -0,0 +1,140 @@
...
+    from aiohttp.web.web_response import Request, StreamResponse
+
+_Handler = Callable[["Request"], Awaitable["StreamResponse"]]
+_Middleware = Callable[["Request", _Handler], Awaitable["StreamResponse"]]
+
+
+@web.middleware
+async def bind_logger(
+    request: "Request", handler: _Handler
+) -> "StreamResponse":
```
rra: Ditto here for `from __future__ import annotations`.

src/safir/middleware.py
```diff
@@ -0,0 +1,140 @@
...
+import structlog
+from aiohttp import web
+
+from safir.logging import response_logger
+
+if TYPE_CHECKING:
+    from aiohttp.web.web_response import Request, StreamResponse
+
+_Handler = Callable[["Request"], Awaitable["StreamResponse"]]
+_Middleware = Callable[["Request", _Handler], Awaitable["StreamResponse"]]
```
rra: I generally move these into the `if TYPE_CHECKING` block since they're only used in type annotations. That also lets you move `Awaitable` and `Callable` into the `if TYPE_CHECKING` block.

tests/test_http.py
```diff
@@ -0,0 +1,22 @@
+"""Tests for the safir.http module.
```
rra: Nit: `pytest` allows either `test_foo.py` or `foo_test.py` naming, but the latter is much friendlier to tab completion, so I usually standardize on the latter for my projects. The semantic hint for humans that these are test files is usually adequately provided by being in the `tests` directory, so not having the `test_` part first isn't missed (at least in my experience).

tests/test_http.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+
+    def create_app():
+        app = web.Application()
+        app.cleanup_ctx.append(init_http_session)
+        return app
+
+    app = create_app()
+    # Need to create the client itself to "start up" the app
+    client = await aiohttp_client(app)  # noqa: F841
```
rra: Nit: Why not drop the assignment so that you don't need the flake8 annotation? Or assert on the type of the `client` return value.

tests/test_http.py
```diff
@@ -0,0 +1,22 @@
...
+from aiohttp import web
+
+from safir.http import init_http_session
+
+
+async def test_init_http_session(aiohttp_client):
+    """Test basic use of init_http_session.
+    """
+
+    def create_app():
```
rra: Missing typing here (`def create_app() -> web.Application:`)

tests/test_middleware.py
```diff
@@ -0,0 +1,50 @@
...
+
+    The procedure here is:
+
+    1. Create an app and bind the middleware to it and configure logging.
+    2. Call an endpoint on the app; that endpoint logs a Hello World message.
+    3. Check the captured log output to ensure a log message with bound
+       context was made.
+    """
+
+    async def handler(request):
```
rra: Missing typing for the return value.

tests/test_middleware.py
```diff
@@ -0,0 +1,50 @@
...
+        logger = request["logger"]
+        logger.info("Hello world")
+        return web.Response(body="OK")
+
+    @dataclass
+    class Configuration:
+        logger_name: str = "testapp"
+
+    def create_app():
```
rra: Missing typing for the return value.

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    documenteer>=0.5,<0.6
+    lsst-sphinx-bootstrap-theme<0.3
+    sphinx-prompt
+    sphinx-automodapi==0.12
+
+[flake8]
+max-line-length = 79
+
+[mypy]
+ignore_missing_imports = True
```
rra: Recommended settings, which would have caught the missing typing noted below:

```
disallow_untyped_defs = True
disallow_incomplete_defs = True
ignore_missing_imports = True
strict_equality = True
warn_redundant_casts = True
warn_unreachable = True
warn_unused_ignores = True
```

setup.py
```diff
@@ -0,0 +1,3 @@
+from setuptools import setup
+
+setup(use_scm_version=True,)
```
rra: This trailing comma looks weird when the call is a single line. Is this a convention from somewhere that I'm not familiar with? (I'm a bit surprised black didn't remove it.)

rra:
Our packages tend to contain some YAML scattered around (Kubernetes manifests, GitHub Pipelines). Have you looked at whether there's a way to get `tox` to run `yamllint` on all YAML files found in the package?

pyproject.toml
jonathansick: aha: https://github.com/asottile/seed-isort-config/issues/15

jonathansick:
@rra I see you're recommending typing the tests, which I hadn't thought about, but makes sense. I've been running mypy like this:

```
[testenv:typing]
description = Run mypy.
commands =
    mypy src/safir
```

I guess it should now be:

```
[testenv:typing]
description = Run mypy.
extras =
    dev
commands =
    mypy src/safir tests
```

rra:
Oh, I missed that. Before, I've always used `mypy .`, but that doesn't work because there is no `src/__init__.py` (nor do I think we want there to be one).

I would use `mypy src/safir tests setup.py`. The chances of a typing issue in our small `setup.py` is low, but generally I try to type every bit of Python code in the project.

setup.py
jonathansick: It seems that black got halfway through reformatting, but didn't drop the trailing comma. It accepts the fixed form.

tests/test_http.py
jonathansick: Good idea. I didn't know that it was an option, but I've switched it now.

jonathansick:
Thanks for all your suggestions @rra I think I've got them implemented.

Re: yaml checking, there's a build-in plugin for pre-commit to do that. See https://pre-commit.com/hooks.html

I'll try it out.

docs/set-up-from-template.rst
```diff
@@ -0,0 +1,184 @@
...
+That's because the Docker build step doesn't credentials for Docker Hub.
+
+To set those credentials, follow GitHub's help page `Creating and storing encrypted secrets <https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets>`__.
+The variables are:
+
+``DOCKER_USERNAME``
+    A Docker Hub username that has access to the lsstsqre organization on Docker Hub
+
+``DOCKER_PASSWORD``
+    The password correspond to ``DOCKER_USERNAME``.
```
rra: We should not recommend storing one's username and password in GitHub. Instead, create a Docker Hub token specific to this application and then store that in GitHub. Unfortunately, Docker Hub has no way to scope tokens, so the token will still have the ability to upload to any repository one has access to, but (a) hopefully they'll add that in the future, and (b) at least we can still revoke the tokens separately.

It's kind of painful that we'll have a proliferation of these tokens owned by random individuals who may leave the project, delete their Docker Hub accounts, or otherwise invalidate the tokens. But that part I think we can live with for now and think about later.

It looks like the Docker integration will come with a later commit. When you do finish that, I think it should assume a `DOCKER_TOKEN` secret and pass in the appropriate things to Docker to use the token. (I think you use some magic string as the username and the token as the password, but I'd have to check the docs again.)

docs/storage-keys.rst
```diff
@@ -0,0 +1,56 @@
...
+Safir uses the ``safir/`` prefix for keys to prevent collisions with your code.
+If your application adds custom data (outside the Safir framework) to the application, consider using a custom prefix to namespace that data.
+
+Safir's application keys
+========================
+
+``safir/config``
+    The conventional key for storing your application's configuration instance.
+
+``safir/httpClient``
```
rra: Why camelCase instead of snake_case here? We seem to be using snake_case elsewhere.

docs/set-up-from-template.rst
jonathansick: Good point. I'll change this and the template to refer to personal access tokens: https://docs.docker.com/docker-hub/access-tokens/

The docs show that you still use your regular username.

rra: Oh, okay, that's perfect then, and it makes sense to keep the username as a secret so that both are stored together. Looks great, thank you!

PR #5, opened 2020-03-04
DM-23761: Add support for Kafka (SQuaRE Events)

jonathansick:
- [ ] Create SSL context for Kafka brokers
- [ ] Create a Producer


jonathansick:
This branch is no longer used by any running app versions, so we can finally retire it.

PR #13, opened 2021-01-14
[DM-28114] Update dev dependencies, fix duplicate logging

rra:
Update dev and pre-commit dependencies and apply corresponding
changes.

Each time safir was configured, it stacked another handler into
the safir logger.  This meant that if logging was repeatedly
configured in the same Python process, all log output to stdout
would be duplicated.  This primarily affected test suites, where
each test would reinitialize logging.  The impact was that failed
tests from files with lots of tests would have log out duplicated
a number of times equal to the number of tests, which made it hard
to find the error or understand the output.

Avoid this by clearing the handlers in the safir logger each time
logging is configured to ensure that only the single, wanted handler
is added.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,6 +11,7 @@ jobs:
         python:
           - 3.7
           - 3.8
+          - 3.9
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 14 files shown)


rra:
This will be easiest to review as individual commits, since most of the commits are general package cleanup unrelated to the feature change.

jonathansick (approved):
Great ideas. Do you want to go ahead and push the `0.1.1` tag from `master` after merging this?

rra:
Yup, will do!

PR #24, opened 2021-06-10
[DM-30689] Optionally add a timestamp to log messages

rra:
Kubernetes tracks the timestamp of each output line internally, so
this is not strictly necessary for our Kubernetes services, but
Argo CD doesn't display that timestamp and other log aggregation may
not necessarily.  Add an optional argument to configure_logging to
allow an application to request a timestamp be added to each log
message.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,9 +58,9 @@ dev =
 
 [flake8]
 max-line-length = 79
-# Ignored rules for Black
-# E203 whitespace before ':'
-ignore = E203
+# E203: whitespace before :, flake8 disagrees with PEP-8
+# W503: line break after binary operator, flake8 disagrees with PEP-8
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -58,9 +58,9 @@ dev =
 
 [flake8]
 max-line-length = 79
-# Ignored rules for Black
-# E203 whitespace before ':'
-ignore = E203
+# E203: whitespace before :, flake8 disagrees with PEP-8
+# W503: line break after binary operator, flake8 disagrees with PEP-8
+ignore = E203, W503
```
rra, line 63: This chunk turned out not to be necessary after refactoring, but it brings safir in line with the roundtable_aiohttp template so I left it.

PR #28, opened 2021-06-22
[DM-28130] Delay AsyncClient creation and allow recreation

rra:
The approach of creating one AsyncClient on startup and destroying
that one client without a way of recreating it on shutdown works
fine for the application when run normally, but fails in test
scenarios that reuse the application.

Instead, defer creation of the AsyncClient until the first time
it's requested, clear it on aclose(), and recreate it if it's asked
for again.  This allows applications to aclose() the client on
shutdown but still work when the same app object is reused for
multiple tests.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Change log
 .. Headline template:
    X.Y.Z (YYYY-MM-DD)
 
-2.0.1 (unreleased)
+2.0.1 (2021-06-24)
 ==================
 
 - Defer creation of ``httpx.AsyncClient`` until the first time it is requested.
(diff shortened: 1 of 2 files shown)


jonathansick (approved):
Very cool, this looks great. Is our testing pattern normally to reuse the app (session scoped fixture?) or is this a special case? I'll have to take a closer look at the testing examples.

rra:
This is the problem where FastAPI has a strong bias towards creating objects at module import time. The standard FastAPI pattern is that you define `app` as a top-level module constant in your `main.py`, which means that there's only one app for the lifetime of a Python process. Due to how pytest works, that means the same app is reused for all tests.

When I first started using FastAPI, I tried to find a way to create the app in a function instead, but it was painful enough that I gave up. I think the specific problem that I ran into dated from before using a router to hold all the routes, so it might be worth another try one of these days.

PR #31, opened 2021-08-09
[DM-31361] Safir improvements from mobu

rra:
- Add the `ErrorModel` to Safir (useful for documentation)
- Mark all dependencies as async so that they won't be run in a thread pool

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,13 @@ Change log
 .. Headline template:
    X.Y.Z (YYYY-MM-DD)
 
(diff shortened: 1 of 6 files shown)


jonathansick (approved):
Nice. One thing that might be useful is an illustrative example or mention in the Error model docstring of how this model should be used with the `responses` argument of the route decorator. I hadn't quite groked this FastAPI feature until I specifically went searching for it in Gafaelfawr.

Second, I think you'll need to add the models specifically to the `automodapi` directive in the doc's `index.rst`?

PR #35, opened 2021-09-03
[DM-31621] Update GitHub Actions workflow

rra:
Don't try to upload documentation on pull requests unless the
pull request is from a ticket branch.  This avoids failed
documentation uploads from dependabot PRs.

Update the CI workflow to match the current recommendation, so that
the workflow is run on pull requests, pre-commit is run with its
own action, and we cache the tox environments.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,11 +1,25 @@
 name: Python CI
 
-on: [push]
+"on":
+  push:
+    branches-ignore:
+      # These should always correspond to pull requests, so ignore them for
+      # the push trigger and let them be triggered by the pull_request
+      # trigger, avoiding running the workflow twice.  This is a minor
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,11 +1,25 @@
 name: Python CI
 
-on: [push]
+"on":
+  push:
+    branches-ignore:
```
jonathansick, line 5: Maybe

```
branches:
  - master
  - main
```

Would be sufficient?

jonathansick (approved):

.github/workflows/ci.yaml
rra: It probably would for Safir but I was trying to standardize on a pattern that would also work for packages that had maintenance branches. But maybe that's overkill?

PR #48, opened 2021-11-09
[DM-32520] Adjust for new httpx

rra:
- Enable redirects by default, restoring the previous behavior
- Depend on the version of httpx adding the new redirect behavior
- Switch from pytest-httpx to respx for testing
- Install `py.typed` so that mypy knows to use type annotations in this library
- Update pre-commit hooks


jonathansick (approved):
Looks great, thanks.

PR #53, opened 2021-12-10
[DM-32941] Change JSON logging for compatibility with Google Log Explorer

rra:
- Log the log level as `severity` instead of `level` when using JSON format
- Log HTTP request information in an `httpRequest` key using the Google-supported fields


jonathansick (approved):
Terrific, thank you. 👍

PR #71, opened 2022-02-09
[DM-33604] Improve handling of X-Forwarded-Proto

rra:
In the XForwardedMiddleware, rather than storing the forwarded
scheme from X-Forwarded-Proto in the request state, simply update
the scope of the request so that Starlette will think the incoming
scheme was the sheme from X-Forwarded-Proto.  Drop the state
variable that was previously set.

This will ensure that other Starlette functions such as url_for
will pick up the correct scheme and thus generate URLs that
correctly use https even if the service is behind an ingress that
terminates TLS.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,12 @@ Change log
 .. Headline template:
    X.Y.Z (YYYY-MM-DD)
 
+3.0.0 (unreleased)
+==================
+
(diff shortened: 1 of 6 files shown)


jonathansick (approved):
The only thing I'm unsure of is making this a 3.0 release rather than a bugfix release?

rra:
My logic was that it breaks backwards compatibility since it no longer sets something in the request state that previously was set (which requires a change in at least Gafaelfawr). But I was sort of on the fence about it, so if you don't think that's a significant enough backwards-compatibility break to warrant a semver change, I'm happy to change the version number.

jonathansick:
Sounds reasonable 👍

PR #72, opened 2022-02-11
[DM-33604] Add FastAPI dependencies for Gafaelfawr auth

rra:
Add new auth_dependency and auth_logger_dependency FastAPI
dependencies for applications protected by Gafaelfawr.  The former
returns the authenticated user from the header set by Gafaelfawr,
and the latter wraps logger_dependency and adds the authenticated
user as a bound context variable.

### DIFF
--- a/docs/gafaelfawr.rst
+++ b/docs/gafaelfawr.rst
@@ -18,7 +18,7 @@ To get that username, use the `~safir.dependencies.gafaelfawr.auth_dependency` F
 
 
    @app.get("/route")
-   async def get_rounte(user: str = Depends(auth_dependench)) -> Dict[str, str]:
+   async def get_rounte(user: str = Depends(auth_dependency)) -> Dict[str, str]:
        # Route implementation using user.
        return {"some": "data"}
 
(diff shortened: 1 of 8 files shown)


docs/gafaelfawr.rst
```diff
@@ -0,0 +1,36 @@
...
+
+.. code-block:: python
+
+   from fastapi import Depends
+
+   from safir.dependencies.gafaelfawr import auth_dependency
+
+
+   @app.get("/route")
+   async def get_rounte(user: str = Depends(auth_dependench)) -> Dict[str, str]:
```
jonathansick, line 21:
```suggestion
   async def get_rounte(user: str = Depends(auth_dependency)) -> Dict[str, str]:
```

docs/gafaelfawr.rst
```diff
@@ -0,0 +1,36 @@
...
+       # Route implementation using user.
+       return {"some": "data"}
+
+If the request does not have the Gafaelfawr header set, it will be rejected by FastAPI with a 422 status code.
```
jonathansick, line 25: Should something be said about testing an app that uses this dependency? Do we have a recommended test client that bakes in this header? Or remind the user that their test client should include this header?

jonathansick (approved):

docs/gafaelfawr.rst
rra: Good idea. I added a new section.

PR #75, opened 2022-02-14
[DM-33604] Move tox configuration to tox.ini

rra:
Unfortunately, tox-docker (which will be used shortly for
testing database code) doesn't support putting the tox
configuration into pyproject.yaml.  We've also switched to a
separate tox.ini in our standard project template, so may as well
make Safir match.


jonathansick (approved):
I do agree. Apparently I was quite optimistic in thinking that TOML/pyproject support was just around the corner!

PR #76, opened 2022-02-15
[DM-33604] Add database support code

rra:
Import and adapt the database dependency from Gafaelfawr and the
database setup and utility functions from Gafaelfawr and vo-cutouts.
Add a new db extra that installs the necessary prerequisites.

### DIFF
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -66,59 +66,6 @@ Note that `~safir.database.initialize_database` returns a `~sqlalchemy.ext.async
 This can be used to perform any further application-specific database initialization that is required, such as adding default table entries.
 Put any such code before the ``await engine.dispose()`` call.
 
-.. _async-db-session:
-
-Creating an async database session
-==================================
-
-(This section describes how to get a database session outside of a FastAPI route handler.
-For details on how to get a database session inside a FastAPI application, see :ref:`fastapi-database-session`.)
-
-To get a new async database connection, use code like the following:
-
-.. code-block:: python
-
-   import structlog
-   from safir.database import create_async_session, create_database_engine
-
-   from .config import config
-
-
-   engine = create_database_engine(config.database_url, config.database_password)
-   session = await create_async_session(engine)
-
-   # ... use the session here ...
-
-   await session.remove()
-   await engine.dispose()
-
-Creating the engine is separate from creating the session so that the engine can be disposed of properly, which ensures the connection pool is closed.
-
-Probing the database connection
--------------------------------
-
-`~safir.database.create_async_session` supports probing the database to ensure that it is accessible and the schema is set up correctly.
-To do this, pass a SQL statement to execute as the ``statement`` argument to `~safir.database.create_async_session`.
-This will be called with ``.limit(1)`` to test the resulting session.
-When ``statement`` is provided, a `structlog`_ logger must also be provided to log any errors when trying to run the statement.
-
-For example:
-
-.. code-block:: python
-
-   import structlog
-   from sqlalchemy.future import select
-
-   from .schema import User
-
-
-   logger = structlog.get_logger(config.logger_name)
-   stmt = select(User)
-   session = await create_async_session(engine, logger, statement=stmt)
-
-If the statement fails, it will be retried up to five times, waiting two seconds between attempts, before raising the underlying exception.
-This is particularly useful for waiting for network or a database proxy to come up when a process has first started.
-
 .. _fastapi-database-session:
 
 Using a database session in request handlers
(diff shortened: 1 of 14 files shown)


docs/database.rst
```diff
@@ -0,0 +1,231 @@
...
+
+To drop and recreate all of the tables, pass the ``reset=True`` option to `~safir.database.initialize_database`.
+
+Note that `~safir.database.initialize_database` returns a `~sqlalchemy.ext.asyncio.AsyncEngine` object for the newly-initialized database.
+This can be used to perform any further application-specific database initialization that is required, such as adding default table entries.
+Put any such code before the ``await engine.dispose()`` call.
+
+.. _async-db-session:
+
+Creating an async database session
```
jonathansick, line 71: So I think the main thing we want to get out of this section is how to write a `check_database` function that's run during application start-up, and then parenthetically say that `create_async_session` can also be used to create a database session outside a request.

So I'd be tempted to reformat this section around that `check_database()` task, and then if we need an explicit section about getting an async session outside a request dependency, maybe put that section after "Using a database session in request handlers", next to "Creating a sync database session".

I think this keeps the most important and actionable docs first.

docs/dev/development.rst
```diff
@@ -63,23 +65,27 @@ To proceed, stage the new modifications and proceed with your Git commit.
...
+   tox -av
 
-To see a listing of test environments, run:
+tox will start a PostgreSQL container, which is required for some tests.
+You can also run pytest_ from the root of the source repository:
 
 .. code-block:: sh
 
-   tox -av
+   pytest
```
jonathansick, line 85: At this point I'm generally never running pytest directly; you can always pass args to pytest, like a specific module with `--`, like

```
tox -e docker -- tests/some/module_test.py
```

src/safir/dependencies/db_session.py
```diff
@@ -0,0 +1,119 @@
...
+
+       await db_session_dependency.initialize(database_url)
+
+    from its startup hook and:
+
+    .. code-block:: python
+
+       await db_session_dependency.aclose()
+
+    from its shutdown hook.
```
jonathansick, line 40: For clarity, I tend not to use blocks within a sentence. Instead, you might want to introduce the `initialize` code sample by saying, "In the startup hook:` and introduce the `aclose` sample by saying "In the shutdown hook:"

jonathansick (approved):

src/safir/dependencies/db_session.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            yield self._session
+
+        # Following the recommendations in the SQLAlchemy documentation, each
+        # session is scoped to a single web request.  However, this all uses
+        # the same async_scoped_session object, so should share an underlying
+        # engine and connection pool.
+        await self._session.remove()
+
+    async def aclose(self) -> None:
```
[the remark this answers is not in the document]
rra: I use `aclose` because it's consistent with [contextlib.aclosing](https://docs.python.org/3/library/contextlib.html#contextlib.aclosing) as distinct from contextlib.closing. Admittedly, none of these dependencies are used with the closing pattern, but it felt like a good general pattern, sort of a level below the corresponding distinction in dunder methods. Python requires dunder methods have different names and doesn't require the close methods have different names, but since there's standard library support for calling them by those names, it felt like a pattern to follow.

docs/database.rst
```diff
@@ -0,0 +1,231 @@
...
+   logger = structlog.get_logger(config.logger_name)
+   engine = await initialize_database(
+       config.database_url,
+       config.database_password,
+       logger,
+       schema=Base.metadata,
+   )
+   await engine.dispose()
+
+(Although not shown here, this must be done inside an ``async`` function.)
```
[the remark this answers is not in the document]
rra: Exactly that is coming in two diffs when I import the decorator function that lets one write Click command functions that are async. This is essentially placeholder text that will be replaced by that upcoming diff.

(That said, I didn't document the testing fixture, and I should do that. There are a lot of things to say about testing, so I think I'll add that as a separate diff, but yes, database initialization has to be covered as well.)

docs/database.rst
rra: I actually don't think you should write a `check_database` call at all, and instead the application startup script should call `app init` before starting the app with uvicorn (which will, as a side effect, test database connectivity), but this should definitely be documented. I'm going to tackle that in a later diff when the Click stuff is introduced, so for this I think I'm going to move the database dependency up as the normal way to interact with the database and rework this section to be clear that this is for non-FastAPI uses of the database, such as cron jobs, background processing, or other supporting code.

docs/dev/development.rst
rra: Good call, I've deleted this bit entirely and instead added an example of running a specific test with `tox`.

PR #77, opened 2022-02-16
[DM-33604] Add datetime helper functions to database layer

rra:
Add datetime_from_db and datetime_to_db helper functions to the
safir.database module, plus tests and documentation.


jonathansick (approved):
This is handy. I wonder if there's a way to use SQLAlchemy type decorator to sort of bake this pattern into our SQLAlchemy models (although since we're using adapters over the database already, it's not a huge deal for most of our code).

Ah yes, SQLAlchemy has an example showing this approach: https://docs.sqlalchemy.org/en/14/core/custom_types.html#store-timezone-aware-timestamps-as-timezone-naive-utc

Up to you if you're interested in pursuing it; I think we can be happy with simple conversion functions like this PR's.

rra:
Oh, thanks, that's a good find! I'm going to merge this in the interest of getting through my patch queue, but I've made a note to come back to this and see if this is a better approach. (I want to test that it still works properly with select clauses.)

PR #79, opened 2022-02-17
[DM-33604] Add run_with_asyncio wrapper and init docs

rra:
We're using variations on this in several packages now to allow
the command-line code to run asyncio functions, so add it to Safir
rather than copy/pasting the code everywhere.

Use this wrapper to document a command-line way of requesting
database initialization, and document the practice of running that
init command during each pod startup.


jonathansick (approved):
Super!

PR #84, opened 2022-02-22
[DM-33662] Refinements to the database support

rra:
After testing with the applications that use this database support, some parts of the API were clunky or didn't allow enough connection sharing to try to optimize test suite execution times. FastAPI 0.74 also introduced a new problem that made the managed transactions mode fail. Make the following changes:

- Change `iniitalize_database` to take the engine as a parameter, rather than create a new engine that the caller is responsible for freeing. The caller should call `create_database_engine` first instead, and let it handle authentication.
- Allow overriding the engine used by `db_session_dependency` in case an application wants to share a connection pool among tests.
- Remove managed transactions mode for `db_session_dependency` since it doesn't work with FastAPI 0.74 and later with uvicorn and asyncpg.

### DIFF
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -45,7 +45,7 @@ For applications using `Click`_ (the recommended way to implement a command-line
    import click
    import structlog
    from safir.asyncio import run_with_asyncio
-   from safir.database import initialize_database
+   from safir.database import create_database_engine, initialize_database
 
    from .config import config
    from .schema import Base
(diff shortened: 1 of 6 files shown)


src/safir/dependencies/db_session.py
```diff
@@ -109,10 +110,29 @@ async def initialize(
...
+            self._session = await create_async_session(self._engine)
+
+    def override_engine(self, engine: AsyncEngine) -> None:
+        """Force the dependency to use the provided engine.
+
+        Intended for testing, this allows the test suite to configure a single
+        database engine and share it across all of the tests, benefiting from
+        connection pooling for a test speed-up.  In the Gafaelfawr test suite,
+        sharing an engine for all tests improved the time it took to run the
+        full test suite by XX%.
```
jonathansick: Do you have the speed-up statistic available? Was it something like 10%?

docs/database.rst
```diff
@@ -158,15 +157,23 @@ Then, any handler that needs a database session can depend on the `~safir.depend
...
+If that transaction is not explicitly ended, `asyncpg`_ may leave it open, which will cause database deadlocks and other problems.
+Due to an as-yet-unexplained interaction with FastAPI 0.74 and later, managing the transaction inside the database session dependency does not work; calling ``await session.commit()`` there, either explicitly or implicitly via a context manager, immediately fails by raising ``asyncio.CancelledError`` and the transaction is not committed or closed.
+The program using the Safir database dependency must therefore explicitly manage transactions itself and ensure that all transactions are committed or rolled back before a handler returns its result.
+
+Generally the easiest way to do this is to manage the transaction in the handler function, as in the ``get_index`` function in the example above.
+Wrap all code that may make database calls in an ``async with session.begin()`` block.
+This will open a transaction, commit the transaction at the end of the block, and roll back the transaction if the block raises an exception.
```
jonathansick, line 176: We might want to re-arrange this slightly for user documentation (keeping in mind that users care more about what they need to do now rather than what was envisioned as being possible in a unreleased version of Safir). I can't do a multi-line patch, but it might go like this:

> You need to manage transactions when using the Safir database dependency. SQLAlchemy will automatically start a transaction if you perform any database operation using a session (including read-only operations). If that transaction is not explicitly ended, asyncpg may leave it open, which will cause database deadlocks and other problems.
> 
> Generally it's best to manage the transaction in the handler function (see the `get_index` example, above). Wrap all code that may make database calls in an `async with session.begin()` block.
This will open a transaction, commit the transaction at the end of the block, and roll back the transaction if the block raises an exception.
> 
> NOTE: Due to an as-yet-unexplained interaction with FastAPI 0.74 and later, managing the transaction inside the database session dependency does not work; calling ``await session.commit()`` there, either explicitly or implicitly via a context manager, immediately fails by raising ``asyncio.CancelledError`` and the transaction is not committed or closed.

(where the "note" is an rst `note` directive).

jonathansick (approved):
Sounds reasonable.

I've included two comments on docs, otherwise looks good.

src/safir/dependencies/db_session.py
rra: It was about 5%, and there's a lot of noise. It was just borderline enough to keep, but I removed the claim of a specific speedup and replaced it with a note that it's probably not worth it unless you have a really long test suite.

docs/database.rst
rra: This is great, thank you!

PR #86, opened 2022-02-24
DM-33627: Retain port setting in safir.database

jonathansick:
This allows non-standard Postgres port settings to be respected. I sometimes use a different port for either the "dev" or "test" containers to avoid conflicts when running tests while also running a dev instance of the app locally.

Also fixes a typo in the "init" CLI function example.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,11 @@ Change log
 .. Headline template:
    X.Y.Z (YYYY-MM-DD)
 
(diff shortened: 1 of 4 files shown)


rra (approved):
Thanks for fixing that. I really dislike the way that the `netloc` thing works when trying to make changes to parts of a parsed URL.

jonathansick:
Yeah, it it's an odd API. Also that `_replace` is a public private method...

PR #95, opened 2022-03-25
[DM-34097] Fix database passwords with special characters

rra:
Be more careful about URL quoting when rewriting database URLs in
safir.database so that passwords containing characters with special
meaning in URLs (@, /) are handled correctly.

urlparse doesn't undo quoting, so use a test database password
containing various special characters for a thorough test.


jonathansick (approved):
Fab. Thanks for sorting this one out.

PR #98, opened 2022-04-19
DM-34473: Add arq FastAPI dependency

jonathansick:
[arq](https://arq-docs.helpmanual.io) is a asyncio distributed Redis queue package. This adds a FastAPI dependency for sending jobs to that queue from a FastAPI app.

This also includes a mock queue, which is useful for testing and isolating the API handler from the worker functions.

This code is ported from lsst-sqre/noteburst where is it currently being used. This code is also used by Times Square as well.

- [x] Add tests
- [x] Add documentation → https://safir.lsst.io/v/DM-34473/arq.html


src/safir/dependencies/arq.py
```diff
@@ -0,0 +1,592 @@
...
+            redis_settings, default_queue_name=default_queue_name
+        )
+        return cls(pool)
+
+    @property
+    def default_queue_name(self) -> str:
+        """Name of the default queue, if the ``_queue_name`` parameter is
+        no set in method calls.
+        """
+        return self._default_queue_name
```
rra, line 382: This doesn't look different from the parent class implementation. I think you can delete it unless I'm missing something.

src/safir/dependencies/arq.py
```diff
@@ -0,0 +1,592 @@
...
+        job = await self._pool.enqueue_job(
+            task_name,
+            *task_args,
+            _queue_name=_queue_name or self.default_queue_name,
+            **task_kwargs,
+        )
+        if job:
+            return await JobMetadata.from_job(job)
+        else:
+            raise JobNotQueued
```
rra, line 400: It just returns `None` on failure with no meaningful error information? That's kind of rude... not sure how you should debug what failed.

src/safir/dependencies/arq.py
```diff
@@ -0,0 +1,592 @@
...
+        self._job_metadata: Dict[str, Dict[str, JobMetadata]] = {
+            self.default_queue_name: {}
+        }
+        self._job_results: Dict[str, Dict[str, JobResult]] = {
+            self.default_queue_name: {}
+        }
+
+    def _resolve_queue_name(self, queue_name: Optional[str]) -> str:
+        return queue_name or self.default_queue_name
+
```
rra, line 438: I'm a huge fan of factoring out common code, but `_queue_name or self.default_queue_name` is only a single character longer than `self._resolve_queue_name(_queue_name)`, so I think this may be overkill?

rra (commented):
Stylistically, I'm not sure I like having all the arq code in the dependency module. Conceptually to me there are two different pieces of code here: an object-oriented convenience API to arq that supports mocking, and a FastAPI dependency that manages an instantiation of that API. I would personally move the former into a `safir.arq` module, and have only the dependency glue in the dependency module.

setup.cfg
```diff
@@ -66,6 +66,8 @@ dev =
     sphinx-prompt
 kubernetes =
     kubernetes_asyncio
+arq =
+    arq==0.23a1
```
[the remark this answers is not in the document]
jonathansick: See commit message prepared for git inspection; will prepare a comment in the file for folks that done use the git history.

src/safir/dependencies/arq.py
rra: Er, maybe this isn't obvious, but I was talking about arq? The implication of this code is that the arq function you're calling might just return `None` with no other errors.

jonathansick: Whoops sorry about that. I didn't catch the context of the code snippet and totally mis-read the comment.

Turns out arq's enqueue method only returns `None` if a job with that ID already exists, so I'll include that in the new exception message. This PR doesn't implement hard-coding job IDs yet though.

PR #101, opened 2022-05-18
[DM-34859] Support pool_pre_ping for create_database_engine

rra:
Support passing pool_pre_ping down to the underlying
create_async_engine API call.  This tells SQLAlchemy to test the
connection whenever a new session is created.

SLAC is seeing a problem where connections are reset when used
during Gafaelfawr login to create a new token while querying the
database to determine if the user is an admin (which is the first
database query that would be done during the login process).  This
seems to be due to a connection timeout, and setting this parameter
should work around the problem.

Setting this unconditionally would be simpler, but it means a
database query every time a session is created, which for some
applications would significantly increase the database traffic and
thus have potential performance issues.  Make it optional so that
the application can choose (and possibly make it configurable in
the application as well).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,13 @@ Change log
 .. Headline template:
    X.Y.Z (YYYY-MM-DD)
 
+3.1.0 (unreleased)
+==================
+
+- Add ``pool_pre_ping`` argument to ``create_database_engine``.
+  If set, this is passed down to SQLAlchemy to enable testing of the database connection each time a connection is checked out of the connection pool.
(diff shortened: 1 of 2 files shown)


rra:
This would have been something we would turn on permanently in some environments where connections were otherwise dying. It would work around quiet termination of the TCP connection by checking and reopening it (similar to the approach I'm taking with bonsai in Gafaelafwr with LDAP connections), at the cost of additional work creating each session.

However, it looks like we've found a change to the PostgreSQL configuration parameters that fix this problem at SLAC, and I really don't want to use pre-pinging if we can avoid it, so I'm going to abandon this change as hopefully unnecessary. We can always revisit it if we see this problem again.

PR #104, opened 2022-06-06
DM-35119: Fixes for current mypy / type checking

jonathansick:
This PR has fixes for two types of mypy checks that came up while developing Safir under current dependencies.

1. mypy reports that `pkg_metadata = cast(Message, metadata(package_name))` is no longer needed in Python 3.10+
2. It seems that starlette's `Request.client` always exists, however, `Request.client.port` can be `None`.

Basically, it seems that these changes are due to improvements in mypy. Are these fixes ok on the basis that adopters of new Safir versions are also adopting the latest mypy and starlette, or would you prefer more defensive changes somehow?

### DIFF
--- a/src/safir/metadata.py
+++ b/src/safir/metadata.py
@@ -3,10 +3,9 @@
 
 from __future__ import annotations
 
-import sys
 from email.message import Message
 from importlib.metadata import metadata
-from typing import Optional, cast
+from typing import Optional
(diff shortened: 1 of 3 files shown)


jonathansick:
False alarm, the problem is that `make init` wasn't dropping the tox working directory as expected.

PR #116, opened 2022-09-08
DM-36156: Drop use of sqlalchemy _IsolationLevel type

jonathansick:
Mypy not longer accepts `sqlalchemy.engine.create._IsolationLevel` as a valid type:

```
src/safir/database.py:131: error: Variable "sqlalchemy.engine.create._IsolationLevel" is not valid as a type src/safir/database.py:131: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

This PR simply removes the use of `_IsolationLevel` for type checking and uses the str type throughout instead. I don't know if there's a more sophisticated solution, but this at least unblocks a clean CI.

### DIFF
--- a/src/safir/database.py
+++ b/src/safir/database.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import asyncio
 import time
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Optional, overload
+from typing import Optional, overload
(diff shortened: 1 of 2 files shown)


rra (approved):
Yup, this looks good. I only did this because the types weren't compatible, but it looks like they fixed that, and plus you can't import the proper type any more.

PR #117, opened 2022-09-08
DM-36156: Refresh Safir packaging and adopt new Rubin user guide theme

jonathansick:
This PR updates Safir with the new Rubin user guide configuration from [Documenteer](https://documenteer.lsst.io/project-guides/guides/index.html) and the new [PyPI package template](https://github.com/lsst/templates/blob/main/project_templates/square_pypi_package).

To make the best use of the new navigation in Rubin user guides, I've re-arranged the documentation here so that we have distinct User guide, API reference, Developer Guide and Change log entries in the top navigation. See the documentation draft at https://safir.lsst.io/v/tickets-DM-36156-packaging. Since Safir and [Documenteer](https://documenteer.lsst.io) are the first packages to use the new documentation theme and architecture, now is a good idea to assess whether we're on the right track with this sort of organizational pattern that might apply to a lot of our Python packages.

Note this PR should be merged after #116


README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Developing Safir
+
+The best way to start contributing to Safir is by cloning this repository creating a virtual environment, and running the init command:
+
+.. code-block:: bash
+
+   git clone https://github.com/lsst-sqre/safir.git
+   cd safir
+   make init
```
rra, line 33: Left-over rST

rra (approved):
This looks good to me! I really like the new layout and the way the top bar is used, and the separation between user guide, API reference, development guide, and change log works for me.

PR #118, opened 2022-09-15
[DM-35879] Support sending Uvicorn logs through structlog

rra:
Add a new function, configure_uvicorn_logging, which overrides the Uvicorn logging configuration to use structlog and adds processors to add the log severity and (for access logs) the details of the request as structlog context.  The context is formatted so that, if logged as JSON, the fields match the expectation of Google's Cloud Logging system.

Unfortunately, there doesn't appear to be an easy way to test this function inside Safir's test suite, since Uvicorn is not friendly to being run as a fixture or in the course of a test function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Headline template:
 X.Y.Z (YYYY-MM-DD)
 -->
 
-## 3.3.0 (unreleased)
+## 3.3.0 (2022-09-15)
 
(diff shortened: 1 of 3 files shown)


jonathansick (approved):
Excellent, thanks.

rra:
Tested by pointing Gafaelfawr at this branch, forcing a Selenium test to fail, and visually inspecting the log messages captured by pytest. The Selenium tests run Gafaelfawr in a subprocess with Uvicorn, avoiding the testing problem, but I'm not sure how to replicate that setup while also using a log capture fixture so that the logs can be checked programmatically.

PR #121, opened 2022-11-18
DM-37046: Add Profile and LogLevel enums

rra:
Add enums with the acceptable values of Profile and LogLevel for configure_logging and configure_uvicorn_logging.  Modify those functions to accept either strings or enums, for backward compatibility.

This change is in preparation for converting the FastAPI template to use pydantic.BaseSettings instead of a dataclass for its Configuration class.  These enums will be used to validate the profile and log_level settings.

This also changes the name of the argument to configure_uvicorn_logging for consistency with configure_logging. Technically this is a backwards-incompatible change, but the recommended way of calling this function doesn't use a named argument, so hopefully this won't break anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Headline template:
 X.Y.Z (YYYY-MM-DD)
 -->
 
+## 3.4.0 (unreleased)
+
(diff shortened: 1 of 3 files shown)


jonathansick (approved):
This looks good, thanks. I'll have to switch my apps' configs over to the new types when this version comes out.

PR #122, opened 2022-11-18
DM-37051: Drop now-unnecessary types from docs

rra:
The current documenteer setup correctly resolves types and doesn't require reiterating them in docstrings in most cases.  Drop them where possible.  Add better docstrings to datetime_from_db and datetime_to_db.

For reasons that I don't understand, structlog.types.EventDict doesn't resolve unquoted but does work in backticks, so add backticks to keep Sphinx happy.

### DIFF
--- a/src/safir/database.py
+++ b/src/safir/database.py
@@ -43,16 +43,16 @@ def _build_database_url(
 
     Parameters
     ----------
-    url : `str`
+    url
         Database connection URL, not including the password.
-    password : `str` or `None`
+    password
         Database connection password.
-    is_async : `bool`
+    is_async
         Whether the resulting URL should be async or not.
 
     Returns
     -------
-    url : `str`
+    url
         The URL including the password.
 
     Raises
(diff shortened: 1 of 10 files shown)


rra:
Sorry about the review request noise. For some reason GitHub was showing no requested reviewers, and then removed you when I tried to add you.

src/safir/database.py
```diff
@@ -91,7 +91,19 @@ def datetime_from_db(time: None) -> None:
 
 
 def datetime_from_db(time: Optional[datetime]) -> Optional[datetime]:
-    """Add the UTC time zone to a naive datetime from the database."""
+    """Add the UTC time zone to a naive datetime from the database.
+
+    Parameters
+    ----------
+    time
+        The naive datetime from the database, or `None`
```
jonathansick, line 99:
```suggestion
        The naive datetime from the database, or `None`.
```

src/safir/database.py
```diff
@@ -91,7 +91,19 @@ def datetime_from_db(time: None) -> None:
...
+
+    Returns
+    -------
+    datetime.datetime or None
```
jonathansick, line 103: I noticed that the `datetime.datetime` gets linked, but the `None` doesn't. I think it's fine tbh, and explicitly marking up the `None` may not be worth it. I guess the rule is that the first "word" gets auto marked up, and the remaining words are treated as text.

src/safir/database.py
```diff
@@ -110,7 +122,21 @@ def datetime_to_db(time: None) -> None:
 
 
 def datetime_to_db(time: Optional[datetime]) -> Optional[datetime]:
-    """Strip time zone for storing a datetime in the database."""
+    """Strip time zone for storing a datetime in the database.
```
jonathansick, line 125: Oh yeah, the treatment of overloads is great.

jonathansick (approved):
Excellent, thanks!

src/safir/database.py
rra: It's a bit more complex in my previous experiments: if you mark up `None`, then `datetime.datetime` doesn't get linked (unless you also mark it up). It's kind of weird.

PR #124, opened 2022-11-28
DM-37051: Add some Pydantic helper functions from Gafaelfawr

rra:
normalize_datetime ensures that input datetime fields are always timezone-aware UTC.  to_camel_case can be used as an alias generator in cases where a camelCase version of the object should be accepted (such as Kubernetes custom resources or YAML files generated from Helm configuration).  validate_exactly_one_of handles validation in cases where exactly one of a set of fields must be set.

to_camel_case and validate_exactly_one_of are also needed by the new lab controller, so will have two uses.  I'm not aware of a second use of normalize_datetime, but I wanted to mention it in SQR-072 to encourage people to standardize on timezone-aware datetime internal to applications.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,8 +21,8 @@ repos:
     rev: v1.12.1
     hooks:
       - id: blacken-docs
-        additional_dependencies: [black==22.8.0]
-        args: [-l, '79', -t, py39]
+        additional_dependencies: [black==22.10.0]
+        args: [-l, '77', -t, py39]
 
   - repo: https://github.com/PyCQA/flake8
     rev: 6.0.0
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -21,8 +21,8 @@ repos:
     rev: v1.12.1
     hooks:
       - id: blacken-docs
-        additional_dependencies: [black==22.8.0]
-        args: [-l, '79', -t, py39]
+        additional_dependencies: [black==22.10.0]
+        args: [-l, '77', -t, py39]
```
rra, line 25: This is a hack, but without it blacken-docs was fighting with flake8 by creating a line longer than 80 characters in a docstring. I think this is a bug in blacken-docs, although maybe I misunderstand it.

jonathansick: I don't have any clue on this. The work around works for me though :)

jonathansick (approved):
Fab. I was hoping these would make it do safir after reading through lsst-sqre/gafaelfawr#633

PR #125, opened 2022-12-02
DM-37192: Enable more mypy checks

rra:
no_implicit_reexport diagnoses importing a symbol from a module that itself imported it from another module, unless that module explicitly lists that symbol in __all__.  local_partial_types is more obscure, preventing mypy from intuiting optional types from assignments across multiple scopes, but it's the default behavior for the mypy daemon and may be the default behavior in the future so it seemed like a good idea to future-proof.

Neither of these produced any new diagnostics in Safir currently. (The first found a bunch of problems in Gafaelfawr.)


jonathansick (approved):
Neat, seems like a good idea.

PR #128, opened 2022-12-14
DM-37291: Add CamelCaseModel helper class

rra:
Add new helper class safir.pydantic.CamelCaseModel, which is identical to pydantic.BaseModel except with configuration added to accept camel-case keys using the safir.pydantic.to_camel_case alias generator.  This is a bit less self-documenting, but is less tedious if there are a lot of classes that need this configuration (such as a complex Kubernetes custom object model).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ X.Y.Z (YYYY-MM-DD)
 
 ## 3.5.0 (unreleased)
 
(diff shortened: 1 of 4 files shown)


jonathansick (approved):
I like this.

My only thought is whether we might want CamelCaseModel to override the implementations of the `dict` and `json` methods to set `by_alias` to `True` by default? I think FastAPI does this by default for endpoint responses, but it'd also be a useful default when using these models to send data to a different services.

rra:
Good idea, done. This was a bit awkward because of typing on those methods, so I did this the hard way and copied a bunch of stuff from Pydantic. It looks like we can do better in Python 3.10.

PR #132, opened 2023-01-17
DM-37589: Add ErrorLocation enum to models

rra:
Add an enum representing the valid first members of the loc array in the ErrorDetail model.  Due to the array (and backward compatibility) the typing can't be enforced, but it's still of use to applications that use Safir and construct custom errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Headline template:
 X.Y.Z (YYYY-MM-DD)
 -->
 
+## 3.6.0 (unreleased)
+
+- Add a `safir.models.ErrorLocation` enum holding valid values for the first element of the `loc` array in `safir.models.ErrorModel`.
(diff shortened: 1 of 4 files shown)


jonathansick (approved):
This is nice to have.

A suggestion on the docs; since they come in as a module docstring these docs come in a kind of odd place (the API page listing). Maybe consider moving the Example content into a new page in the user guide called "Handling errors with HTTPException" or something like that.

rra:
Yeah, this was just me being lazy because I previously hadn't moved the docs into the manual. I'll do that.

rra:
I went ahead and re-requested a review so that you could look over the docs. Thanks!

rra:
The build failure is https://github.com/tox-dev/sphinx-autodoc-typehints/issues/305.

jonathansick (approved):
Excellent, the new doc pages help a lot.

PR #134, opened 2023-01-26
DM-37589: Add datetime utility functions

rra:
These are fairly small, but I find myself copying and pasting them between projects anyway.  current_datetime provides a normalized datetime for the current time.  isodatetime and parse_isodatetime handle the most useful version of ISO 8601 date representation. normalize_isodatetime does the same for input to Pydantic models in cases where accepting all the variations that Pydantic accepts isn't desirable.


jonathansick (approved):
Looks great, thanks.

PR #137, opened 2023-02-01
DM-37760: Change the change log format

rra:
Adopt the change log format used by Gafaelfawr, which divides change log entries into four standard sections. Document the new format and change the existing entries to match.

Change some references from `master` to `main`.


jonathansick (approved):
Nice update, thanks

PR #141, opened 2023-03-06
DM-38170: Fix Uvicorn logging, add Uvicorn testing function

rra:
The structlog Uvicorn logging stack was missing the processor to expand exception backtraces, so only the stringification of the raw exc_info was included. Add the appropriate processor. Also send the access logs to standard output so that they can be easily separated from the other logs (sent to standard error) if desired.

Testing this (and undoing my previous laziness about testing the Uvicorn logging setup) required adding code to spawn a FastAPI app in an external Uvicorn process. Do this by importing the utility code from Gafaelfawr, which uses a similar pattern to set up Selenium testing.


jonathansick (approved):
This looks great, thanks.

PR #143, opened 2023-03-10
DM-38272: Add a Slack webhook client and error reporting

rra:
Add a Slack client and associated models that allow posting structured messages to Slack using the same formatting we currently use for mobu. The models handle truncation and validation of the messages according to the Slack API limitations.

Add a SlackException class that knows how to convert itself to a structured Slack message for rich error reporting. This can be used as a base class for application exceptions that track additional application-specific context data about the source of an exception.

Add a custom FastAPI route class that reports all uncaught exceptions to Slack, along with a SlackIgnoredException base class that excludes exceptions from this reporting. If this class is not configured, it silently does nothing and still re-raises the exception, so configuring a Slack webhook is not required.

Add support for mocking a Slack webhook, taken from Gafaelfawr.

The Slack exception reporting prompted two minor additions to the Safir datetime library: a new format_datetime_for_logging function taken from Gafaelfawr that formats a datetime object in a more human-friendly format than isodatetime, and the optional ability for current_datetime to return a datetime with microseconds.


docs/user-guide/index.rst
```diff
@@ -27,3 +27,4 @@ User guide
    fastapi-errors
    datetime
    uvicorn
+   slack
```
jonathansick, line 30: Consider renaming to `slack-webhooks`

src/safir/slack.py
```diff
@@ -0,0 +1,488 @@
...
+            extra_needed = len(heading)
+            max_length = self.max_formatted_length - extra_needed
+            body = _format_and_truncate_at_end(self.text, max_length)
+        return {"type": "mrkdwn", "text": heading + body, "verbatim": True}
+
+
+class SlackAttachment(SlackField):
+    """An attachment in a Slack message with a heading.
+
+    Intended for use in the ``attachments`` portion of a Block Kit message. If
```
jonathansick, line 95: Do we want to mention anywhere that attachments are considered legacy by Slack, so they might be made less visible over time?

src/safir/slack.py
```diff
@@ -0,0 +1,488 @@
...
+    "SlackException",
+    "SlackIgnoredException",
+    "SlackField",
+    "SlackIgnoredException",
+    "SlackMessage",
+    "SlackRouteErrorHandler",
+]
+
+
+class SlackField(BaseModel):
```
jonathansick, line 30: Would it be cleaner to have a SlackTextField and a SlackCodeField?

jonathansick (changes requested):
This looks really awesome for our app error and event reporting. My big-picture critique is that this new module and classes are sitting on the "slack" and "SlackClient" names, while in fact they're kind of specific to the Slack incoming webhook API. For the work we're doing with Squarebot and it's backends, we'll be adding a client for the [Slack Events API](https://api.slack.com/apis/connections/events-api). I imagine there'll be a `SlackEventsClient` that will work with the [Slack Event Types](https://api.slack.com/events). And while Squarebot itself will be the only app that directly receives and parses requests from Slack, the other apps will use the Slack event models and send requests through the [Slack Web API](https://api.slack.com/web) (either also as part of the SlackEventsClient, or maybe as part of a separate SlackWebApiClient. In Safir I'll be adding Pydantic models for the [Event](https://api.slack.com/events) payloads, and also for [interaction payloads](https://api.slack.com/reference/interaction-payloads) and of course BlockKit fields. So definitely the web API clients and the webhook client can share the same fundamental BlockKit modelling code.

So I guess my main request is that we adjust the naming of the SlackClient and the MockSlack to associate with the webhooks API. It might be ok to still keep all these in the `safir.slack` namespace; definitely I can add the other client(s) to the same namespace and we might end up using private modules within `safir.slack` to organize the code. Because of the number of models, I'm thinking we'd probably want to create a `safir.slack.events` module for the event payloads, and a `safir.slack.blockkit` module for block kit modules.

rra:
This is an excellent suggestion, thank you! Now implemented, along with the other things you mentioned.

jonathansick (approved):
Awesome, thanks for doing this. I look forward to using this everywhere and the other Slack app functionality should slot in well.

PR #144, opened 2023-03-31
DM-38466: Require Python 3.11 and use it for type annotations

rra:
- Set Python 3.11 to the minimum version and update GitHub Actions and other configuration accordingly
- Enable additional mypy checks for Pydantic models and adjust the code accordingly (fixing a multiple inheritance problem with maximum lengths of Slack messages)
- Import types from `collections.abc` instead of `typing` when possible
- Use `dict`, `list`, and `tuple` in types instead of their `typing` equivalents
- Use the `|` syntax instead of `Union`, and instead of `Optional` in cases where the parameter is not optional but may be `None`
- Use `Self` as the return type for class method constructors
- Redo how the `dict` and `json` overrides in `CamelCaseModel` are typed: use a decorator to copy the type signature from the parent method, allowing the override to be written using `kwargs` and therefore be agnostic to changes to the parent type signature without losing all typing information

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
 
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.0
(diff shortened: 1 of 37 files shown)


jonathansick (approved):
Thanks for this, this is great. I hadn't realized all those types from `typing` will likely be dropped someday so I'll have to get used to the new types. Also good to know about the nuance of Optional vs `| None`.

PR #146, opened 2023-03-31
DM-38491: Add PydanticRedisStorage and EncryptedPydanticRedisStorage

jonathansick:
These classes are ported from Gafaelfawr, and refactored to allow both unencrypted and encrypted storage. A draft of the docs is at https://safir.lsst.io/v/DM-38491/user-guide/pydantic-redis.html and API docs are at https://safir.lsst.io/v/DM-38491/api/safir.redis.PydanticRedisStorage.html#safir.redis.PydanticRedisStorage

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -54,6 +54,10 @@ API reference
 .. automodapi:: safir.pydantic
    :include-all-objects:
 
+.. automodapi:: safir.pydanticredis
+   :include-all-objects:
+   :inherited-members:
+
 .. automodapi:: safir.slack.blockkit
 
 .. automodapi:: safir.slack.webhook
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -54,7 +54,7 @@ API reference
 .. automodapi:: safir.pydantic
    :include-all-objects:
 
-.. automodapi:: safir.pydanticredis
+.. automodapi:: safir.redis
    :include-all-objects:
    :inherited-members:
 
--- a/src/safir/redis.py
+++ b/src/safir/redis.py
@@ -5,7 +5,13 @@ from __future__ import annotations
 from collections.abc import AsyncIterator
 from typing import Generic, Optional, TypeVar
 
-import redis.asyncio as redis
+try:
+    import redis.asyncio as redis
+except ImportError:
+    raise ImportError(
+        "The safir.redis module requires the redis extra. "
+        "Install it with `pip install safir[redis]`."
+    )
 from cryptography.fernet import Fernet
 from pydantic import BaseModel
 
--- a/src/safir/redis.py
+++ b/src/safir/redis.py
@@ -1,4 +1,4 @@
-"""Storage for Pydantic models in Redis."""
+"""Redis database support."""
 
 from __future__ import annotations
 
--- a/src/safir/redis.py
+++ b/src/safir/redis.py
@@ -9,6 +9,8 @@ import redis.asyncio as redis
 from cryptography.fernet import Fernet
 from pydantic import BaseModel
 
+from .slack.blockkit import SlackException, SlackMessage, SlackTextField
+
 __all__ = [
     "PydanticRedisStorage",
     "EncryptedPydanticRedisStorage",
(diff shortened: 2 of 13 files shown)


rra (approved):
This looks great. Thank you so much for doing this! Just a few very minor nits. Looking forward to moving this code out of Gafaelfawr!

docs/api.rst
```diff
@@ -54,6 +54,10 @@ API reference
 .. automodapi:: safir.pydantic
    :include-all-objects:
 
+.. automodapi:: safir.pydanticredis
```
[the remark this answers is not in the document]
jonathansick: I like this idea. There's not a lot the pydantic-related implemented would add to the safir.redis namespace, so it'd probably work well alongside whatever else we might add in the future.

jonathansick:
@rra I went ahead and moved the module to `safir.redis`. I also changed the exception to a SlackException. Does the change and additional tests in https://github.com/lsst-sqre/safir/pull/146/commits/6f879a7c4bf9d224fa92e70135991c04b4cf582a look reasonable to you?

src/safir/redis.py
```diff
@@ -0,0 +1,235 @@
...
+import redis.asyncio as redis
+from cryptography.fernet import Fernet
+from pydantic import BaseModel
+
+from .slack.blockkit import SlackException, SlackMessage, SlackTextField
+
+__all__ = [
+    "PydanticRedisStorage",
+    "EncryptedPydanticRedisStorage",
+    "S",
```
rra, line 17: I don't think you want to export this. It's only used by mypy and thus is module local. (Unless this helps the API documentation look nicer, but it looks like you're still excluding it from Sphinx warnings, so presumably Sphinx doesn't yet understand TypeVars.)

jonathansick:
I also want to double check that adding `redis` as a default dependency, not an extra, is ok with you. We have extras for things like `sqlalchemy`; I'm not sure whether you'd judge it a burden to install by default or not?

src/safir/redis.py
```diff
@@ -0,0 +1,235 @@
...
+        data = await self._redis.get(full_key)
+        if not data:
+            return None
+
+        try:
+            return self._deserialize(data)
+        except Exception as e:
+            error = f"{type(e).__name__}: {str(e)}"
+            msg = f"Cannot deserialize data for key {full_key}: {error}"
+            raise DeserializeError(msg, key=full_key) from e
```
rra, line 123: So here this would look like:
```suggestion
            error = f"{type(e).__name__}: {str(e)}"
            msg = f"Cannot deserialize data for key {full_key}"
            raise DeserializeError(msg, key=full_key, error=error) from e
```

src/safir/redis.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+
+    def __init__(self, msg: str, key: str) -> None:
+        super().__init__(msg)
+        self.key = key
+
+    def to_slack(self) -> SlackMessage:
+        message = super().to_slack()
+        message.fields.append(SlackTextField(heading="Key", text=self.key))
+        return message
```
rra, line 34: Because this is Pydantic, which uses newlines and indentation in its exception messages that will be formatted oddly in the main part of the Slack message, I think this would read better in Slack if you did something like this instead:
```suggestion
    def __init__(self, msg: str, key: str, error: str) -> None:
        super().__init__(msg)
        self.key = key
        self.error = error

    def to_slack(self) -> SlackMessage:
        message = super().to_slack()
        message.fields.append(SlackTextField(heading="Key", text=self.key))
        message.blocks.append(SlackCodeBlock(heading="Error", code=self.error))
        return message
```
Then leave the error out of the main exception message. (In the backtrace it will still be captured because the underlying exception will also be logged directly.)

(Maybe this is overkill; if it doesn't sound like a good idea to you, feel free to drop the idea.)

rra (approved):
This generally looks great, just one formatting tweak that I think will improve the eventual Slack rendering.

rra:
Oh, that's a good question on adding Redis as a default dependency. I think I would lean towards making it an option instead. It's not that heavy-weight of a dependency, but I also think most Safir applications won't use Redis (although I dunno, at the rate we're going with Redis deployments, maybe I'm wrong about that!).

PR #151, opened 2023-04-03
DM-38490: Support for building GitHub Apps

jonathansick:
This collects GitHub App infrastructure from our existing GitHub Apps like Semaphore and Times Square. It includes a factory for creating GitHub API clients authenticated either as the app or an app installation, and also includes Pydantic models for key GitHub API resources.

Documentation:

- [Guides](https://safir.lsst.io/v/DM-38490/user-guide/github-apps/index.html)
- [API](https://safir.lsst.io/v/DM-38490/api.html#module-safir.github)


jonathansick:
@rra I didn't really know the right approach for testing the GitHubAppClientFactory; I'm curious to know what you'd do for that.

docs/user-guide/github-apps/handling-webhooks.rst
```diff
@@ -0,0 +1,174 @@
...
+       webhook_secret = config.github_webhook_secret.get_secret_value()
+       event = Event.from_http(request.headers, body, secret=webhook_secret)
+
+       # Bind the X-GitHub-Delivery header to the logger context; this identifies
+       # the webhook request in GitHub's API and UI for diagnostics
+       logger = logger.bind(github_delivery=event.delivery_id)
+
+       logger.debug("Received GitHub webhook", payload=event.data)
+       # Give GitHub some time to reach internal consistency.
+       await asyncio.sleep(1)
```
rra, line 49: This seems worth a mention in the documentation. How did you arrive at that delay? How would one know if the delay is too short?

src/safir/github/_client.py
```diff
@@ -0,0 +1,132 @@
...
+        str
+            The JWT token.
+        """
+        return gidgethub.apps.get_jwt(
+            app_id=self.app_id, private_key=self.app_key
+        )
+
+    def _create_client(
+        self, *, oauth_token: Optional[str] = None
+    ) -> gidgethub.httpx.GitHubAPI:
```
rra, line 53: Nit: Here and elsewhere, given that you're importing `GitHubAPI` directly, I would use it directly in the type signature as well, instead of the fully-qualified type name.

src/safir/github/models.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+
+
+class GitHubUserModel(BaseModel):
+    """A Pydantic model for the ``user`` field found in GitHub API resources.
+
+    This contains brief (public) info about a user.
+    """
+
+    login: str = Field(title="Login name", description="GitHub username.")
```
rra, line 53: It's a shame that the Sphinx Pydantic docs extension seems to mostly throw away the `title` so you have to repeat yourself in `description`, but looking at the rendered docs, this seems like the right approach.

src/safir/github/models.py
```diff
@@ -0,0 +1,372 @@
...
+class GitHubPullRequestModel(BaseModel):
+    """A Pydantic model for a GitHub Pull Request.
+
+    This is also the ``pull_request`` field inside the
+    `~safir.github.webhooks.GitHubPullRequestEventModel`.
+
+    https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request
+    """
+
+    html_url: HttpUrl = Field(description="Web URL of the PR.")
```
rra, line 150: It feels weird to have `description` and not `title`, but given that we're never using these models in an OpenAPI schema (where that would matter), it doesn't seem to matter. I admit I would be tempted to add `title` just in case for the future, but I would probably just be making tedium for myself.

src/safir/github/webhooks.py
```diff
@@ -0,0 +1,343 @@
...
+    "GitHubCheckRunEventModel",
+]
+
+
+class GitHubAppInstallationModel(BaseModel):
+    """A Pydantic model for the ``installation`` field found in webhook
+    payloads for GitHub Apps.
+    """
+
+    id: str = Field(description="The installation ID.")
```
rra, line 38: For these, I probably would add `title` since they're going to go into the OpenAPI documentation of the app that defines this webhook, but it's likely I'm being too pedantic.

src/safir/github/webhooks.py
```diff
@@ -0,0 +1,343 @@
...
+        """The name of the repository owner."""
+        return self.full_name.split("/")[0]
+
+
+class GitHubAppInstallationEventAction(str, Enum):
+    """The action performed on an GitHub App ``installation`` webhook
+    (`GitHubAppInstallationEventModel`).
+    """
+
+    #: Someone installed a GitHub App on a user or organization account.
```
rra, line 101: Out of curiosity, how do you decide between this and the docstring syntax? Is one or the other better? Should we converge on one for our coding style? I've been using the docstring approach, but I'm happy to change.

rra (approved):
This looks great. The documentation is fantastic. Just some really minor comments.

If you really wanted to test the client factory, I think it would involve mocking out the GitHub API and then checking what requests it made, but you'd mostly be testing Gidgethub, which doesn't seem like a great use of time and resources. I guess the other option would be to patch a mock in for the Gidgethub API and just check that you're making the API calls you expect to make, but this code is simple enough that I don't think that sort of test does anything useful other than be fragile on future changes. So in this case I too would talk myself into not writing a test suite for that particular class, since its implementation is so straightforward and is just a wrapper around another module.

docs/user-guide/github-apps/handling-webhooks.rst
jonathansick: I think I got that trick from the Gidgethub docs, who I'm sure got that from experience. I'll mention it in the docs.

src/safir/github/models.py
jonathansick: Yeah, specifically this is the [autodoc_pydantic](https://autodoc-pydantic.readthedocs.io/en/stable/) extension, where the description is the key documentation field. It's definitely a different treatment from redoc/FastAPI. So the tension will mostly come up if we need to both document a Pydantic model as a Pydantic model and as an OpenAPI schema.

src/safir/github/webhooks.py
jonathansick: This was mostly a shot at getting enum documentation to work in the first place. I'll switch back to the docstring syntax to see if that works.

src/safir/github/webhooks.py
jonathansick: In practice I don't have these appear in the OpenAPI docs for the webhook endpoint because I defer Pydantic deserialization to the webhook handler after the Gidgethub's verification and routing steps, rather than having FastAPI deserialize the data into any one model.

rra: Oh! Yes, that was the piece I was missing. I'm good with omitting `title` in that case, although I'm guessing that when I add that plugin to the Gafaelfawr documentation build for better model documentation, I'm going to end up having to write a bit of duplicate documentation.

PR #152, opened 2023-04-11
DM-38533: Merge Kubernetes API mock with lab controller work

rra:
In order to test the new Nublado lab controller, we had to make multiple improvements to the Kubernetes mock and fix a few places where it was incompatible with the kubernetes_asyncio API. Merge those changes back into the Safir version and add additional documentation.

This also adds a new strip_none utility function for removing None key values from possibly-nested complex data structures, which makes it easier to compare Kubernetes objects in serialized form.

Add basic API documentation to all the supported methods of the Kubernetes mock. For reasons that I don't understand, this requires fully-qualified types to avoid Sphinx warnings, even though unqualified types have been working everywhere else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,25 @@ X.Y.Z (YYYY-MM-DD)
 ### Backwards-incompatible changes
 
 - Safir now requires a minimum Python version of 3.11.
+- Custom Kubernetes objects are no longer copied before being stored in the Kubernetes mock by `create_namespaced_custom_object`, and no longer automatically get a `metadata.uid` value. This brings handling of custom objects in line with the behavior of other mocked `create_*` and `replace_*` APIs.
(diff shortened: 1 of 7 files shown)


jonathansick (approved):
Looks great. I look forward to trying this out with the Schema Registry Operator. Not sure about the fully qualified names for the docs — perhaps `kubernetes_asyncio` is doing something clever?

rra:
I think what might be happening is that unqualified names work fine if Sphinx is able to resolve them, but if it's not, it reports an error for the unqualified name which is therefore not suppressed. So I have to fully-qualify the names so that the warning will be suppressed properly. It makes sense in retrospect, it was just confusing me at the time.

All the `kubernetes_asyncio` functions do have docstrings (in fact, the docstrings control the behavior of the library via docstring introspection), but sadly there's no Sphinx-generated documentation site that I know of.

PR #154, opened 2023-04-12
DM-38660: Add SlackWebException for HTTPX errors

rra:
Provide a SlackWebException base exception class, derived from SlackException, which understands how to extract useful information from HTTPX exceptions so that it can be reported to Slack. It also improves the stringification of HTTPX exceptions, which normally may not include the failing URL in the exception stringification.

Unfortunately, automodapi doesn't generate documentation for class or instance methods of exceptions, so the references to from_exception in the docs can't be links to the documentation.


rra:
https://github.com/astropy/sphinx-automodapi/issues/162 is the issue for missing documentation for exceptions.

jonathansick (approved):

PR #155, opened 2023-04-12
DM-38660: Add support for generating structured errors

rra:
Add an exception base class, ClientRequestError, and a corresponding exception handler that formats those exceptions into a JSON error message compatible with safir.models.ErrorModel and FastAPI's internal errors. Support error location tracking either via constructor arguments or via catching the exception, setting some attributes, and re-raising it.


jonathansick (approved):
This looks great. I look forward to trying it out.

PR #158, opened 2023-04-14
DM-38414: Add nested docstring for validate_exactly_one_of

rra:
When used with autodoc_pydantic, generating a summary of validators expects a docstring on the returned validator generated by validate_exactly_one_of. We can't make that great of a docstring, but put in something to avoid the warnings.


rra:
This unfortunately didn't fix the warnings, so there's no point.

PR #166, opened 2023-05-08
tickets/DM-38972 : add fileserver object support

athornton:


src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+                    )
+                    return V1IngressList(kind="Ingress", items=[ingress])
+                except ApiException:
+                    return V1IngressList(kind="Ingress", items=[])
+            else:
+                ingresss = []
+                if "Ingress" in self._objects[namespace]:
+                    for obj in self._objects[namespace]["Ingress"].values():
+                        ingresss.append(obj)
+                return V1IngressList(kind="Ingress", items=ingresss)
```
rra, line 991: `label_selector` should be supported for the non-watch case as well for consistency.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        Parameters
+        ----------
+        name
+            Name of ingress object.
+        namespace
+            Namespace of secret object.
+        body
+            Patches to apply. Only patches with ``op`` of ``replace`` are
+            supported, and only with ``path`` of
+            ``/status/load_balancer/ingress``.
```
rra, line 1020: This should be ``/status/loadBalancer/ingress`` since patches use the Kubernetes API field names, not the Python renamings of them.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+
+        Returns
+        -------
+        kubernetes_asyncio.client.V1Ingress
```
rra, line 1024: Missing a description of the return value.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        kubernetes_asyncio.client.ApiException
+            Raised with 404 status if the ingress does not exist.
+        AssertionError
+            Raised if any other type of patch was provided.
+        """
+        self._maybe_error("patch_namespaced_ingress", name, namespace)
+        ingress = copy.deepcopy(self._get_object(namespace, "Ingress", name))
+        for change in body:
+            assert change["op"] == "replace"
+            assert change["path"] == "/status/load_balancer/ingress"
```
rra, line 1037:
```suggestion
            assert change["path"] == "/status/loadBalancer/ingress"
```

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+
+    # JOB API
+
+    async def create_namespaced_job(self, namespace: str, body: V1Job) -> None:
+        """Create a job object.
+
+        This will in turn create a pod object.  This pod object will have a
+        label "job-name", which will be set to the job's name.  If the mock's
+        ``initial_pod_phase`` is ``Running``, then the Job status will
+        have ``active`` set to 1.
```
rra, line 1053:
```suggestion
        A pod corresponding to this job will also be created. The pod will
        have a label ``job-name`` set to the name of the Job object. If
        ``initial_pod_phase`` on the mock is set to ``Running``, the
        ``status.active`` field of the job will be set to 1.
```

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        # Pretend to spawn a pod
+        # Use our metadata for spawned pod metadata if we weren't given it.
+        template = body.spec.template
+        podmd = template.metadata
+        if podmd is None:
+            podmd = V1ObjectMeta(
+                namespace=namespace,
+                labels=body.metadata.labels.copy(),
+                annotations=body.metadata.annotations.copy(),
+            )
```
rra, line 1080: Is copying metadata from the job what Kubernetes does, or does it require that the metadata part of the template be set? I'm used to always setting this in deployments, so I'm not sure what the documented behavior is.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        podmd.name = name + "-abcde"  # Repeatable for testing
```
rra, line 1081: This should also be documented in the docstring, I think.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        podmd.labels["job-name"] = name
+        await self.create_namespaced_pod(
+            namespace, V1Pod(metadata=podmd, spec=template.spec)
+        )
+        pod = await self.read_namespaced_pod(
+            name=podmd.name, namespace=namespace
+        )
```
rra, line 1088: The created pod is modified in place (intentionally), so you can just do:
```suggestion
        pod = V1Pod(metadata=podmd, spec=template.spec)
        await self.create_namespaced_pod(namespace, pod)
```

rra, line 1088: The created pod is modified in place (intentionally), so you can just do:
```suggestion
        pod = V1Pod(metadata=podmd, spec=template.spec)
        await self.create_namespaced_pod(namespace, pod)
```

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+        if pod.status.phase == "Running":
+            body.status = V1JobStatus(active=1)
+
+    async def delete_namespaced_job(
+        self, name: str, namespace: str, propagation_policy: str = "Foreground"
```
rra, line 1093: If you want to force all of our code to provide the value `Foreground`, then this shouldn't have a default and you should assert that the correct value is passed in. Otherwise, the parameter shouldn't be present unless it is supported and causes a behavior change, which it doesn't look like it does. We don't want to just silently accept other values that we don't support.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +824,426 @@ async def list_namespaced_event(
...
+    ) -> V1Status:
+        """Delete a job object.  Will also propagate to pods.
```
rra, line 1095:
```suggestion
        """Delete a job object.
        
        Will also propagate to pods.
```

src/safir/testing/kubernetes.py
```diff
@@ -44,6 +52,24 @@
...
 
+def _stringify_selector_dict(selector: dict[str, str]) -> str:
+    """This and the following function transmute selector labels between
+    dicts and strings.  We don't even try to do safe quoting."""
+    sstrings: list[str] = []
+    for key in selector:
+        sstrings.append(f"{key}=={selector[key]}")
+    return ",".join(sstrings)
+
+
```
rra, line 63: Unused code, can be deleted.

src/safir/testing/kubernetes.py
```diff
@@ -44,6 +52,24 @@
...
+def _dictify_selector_str(selector_str: str) -> dict[str, str]:
+    sstrings = selector_str.split(",")
+    retval: dict[str, str] = {}
+    for sel in sstrings:
+        (key, val) = sel.split("==", 1)
+        retval[key] = val
+    return retval
```
rra, line 70: While dictify is more specific, it's an awkward term. I think this is just parsing, so I'd call this `_parse_label_selector`. Would be good to add a proper docstring and comment that it only supports equality (and only `==`, not `=`, although ideally we would support both since Kubernetes does). Kubernetes calls each stanza a "requirement."
```suggestion
def _parse_label_selector(label_selector: str) -> dict[str, str]:
    """Parse a label selector.
    
    Only equality is supported (with ``=`` or ``==``).
    
    Parameters
    ----------
    label_selector
        Label selector string to parse.
        
    Returns
    ------
    dict of str to str
        Dictionary of required labels to their required values.
        
    Raises
    ------
    AssertionError
        Raised if the label selector string is not a supported syntax.
    """
    result = {}
    for requirement in label_selector.split(","):
        match = re.match(r"([^!=]+)==?(.*)", requirement)
        assert match and match.group(1) and match.group(2)
        result[match.group(1)] = match.group(2)
    return result
```

rra, line 70: While dictify is more specific, it's an awkward term. I think this is just parsing, so I'd call this `_parse_label_selector`. Would be good to add a proper docstring and comment that it only supports equality (and only `==`, not `=`, although ideally we would support both since Kubernetes does). Kubernetes calls each stanza a "requirement."
```suggestion
def _parse_label_selector(label_selector: str) -> dict[str, str]:
    """Parse a label selector.
    
    Only equality is supported (with ``=`` or ``==``).
    
    Parameters
    ----------
    label_selector
        Label selector string to parse.
        
    Returns
    -------
    dict of str to str
        Dictionary of required labels to their required values.
        
    Raises
    ------
    AssertionError
        Raised if the label selector string is not a supported syntax.
    """
    result = {}
    for requirement in label_selector.split(","):
        match = re.match(r"([^!=]+)==?(.*)", requirement)
        assert match and match.group(1) and match.group(2)
        result[match.group(1)] = match.group(2)
    return result
```

src/safir/testing/kubernetes.py
```diff
@@ -1010,8 +1518,17 @@ async def list_namespaced_pod(
...
+                if "Pod" in self._objects[namespace]:
+                    for obj in self._objects[namespace]["Pod"].values():
+                        if label_dict:
+                            pod_labels = obj.metadata.labels
+                            for lbl in label_dict:
+                                if (
+                                    lbl not in pod_labels
+                                    or label_dict[lbl] != pod_labels[lbl]
+                                ):
+                                    continue
```
rra, line 1530: Same problem here that this code doesn't do anything. It's probably worth writing a utility function that checks a labels dict against a dict of label requirements, since this same pattern is in four places. (Maybe a utility function that handles both field selectors and label selectors, actually, and incorporates the parsing.)

src/safir/testing/kubernetes.py
athornton: I'm not sure either.

athornton: I've done a little investigation.  It doesn't copy labels.  It does create a Metadata object with the namespace, a job-name label, and generateName with the job name prepended to a hyphen as a prefix.  That's what I'm doing now.

src/safir/testing/kubernetes.py
rra: I'm not seeing where `label_selector` is supported for the non-watch case.

src/safir/testing/kubernetes.py
```diff
@@ -742,6 +861,435 @@ async def list_namespaced_event(
...
+                if "Job" in self._objects[namespace]:
+                    for obj in self._objects[namespace]["Job"].values():
+                        if label_dict:
+                            job_labels = obj.metadata.labels
+                            for lbl in label_dict:
+                                if (
+                                    lbl not in job_labels
+                                    or label_dict[lbl] != job_labels[lbl]
+                                ):
+                                    continue
```
rra, line 1253: This still has that bug where the for loop doesn't do anything because the continue is within the loop. I think you want to call your new helper function.

src/safir/testing/kubernetes.py
```diff
@@ -1296,6 +1854,144 @@ async def create_namespaced_service(
...
+                    )
+                    return V1ServiceList(kind="Service", items=[service])
+                except ApiException:
+                    return V1ServiceList(kind="Service", items=[])
+            else:
+                services = []
+                if "Service" in self._objects[namespace]:
+                    for obj in self._objects[namespace]["Service"].values():
+                        services.append(obj)
+                return V1ServiceList(kind="Service", items=services)
```
rra, line 1955: Should support label selector in the not-watch case.

rra (commented):
Just a few remaining issues around label selector, I think. Otherwise looks good.

src/safir/testing/kubernetes.py
athornton: I'm wondering if I have unpushed changes here somewhere.

src/safir/testing/kubernetes.py
athornton: Yeah, missed this one too.

src/safir/testing/kubernetes.py
athornton: Or maybe I was just sloppy.

athornton:
Anyway, think I've got it now.

rra (approved):
Yup, looks good. Thanks!

PR #167, opened 2023-05-16
DM-38425: Extract more data from httpx exceptions

rra:
Previously, SlackWebException assumed that if an httpx exception inherited from RequestError, it would have a request attribute, and if it didn't, it wouldn't. This turns out to be incorrect; some RequestError exceptions may not have a request for various reasons, and some that do not inherit from RequestError do.

Adjust the exception parsing to take this into account by attempting to extract the request and falling back on the non-request handling on any error.

Fix one place where SlackTextField was used instead of SlackTextBlock for something added to the message blocks. (The different length restriction is unlikely to matter in practice, but this is more correct.)

### DIFF
--- a/src/safir/slack/blockkit.py
+++ b/src/safir/slack/blockkit.py
@@ -6,7 +6,7 @@ from abc import ABCMeta, abstractmethod
 from datetime import datetime
 from typing import Any, ClassVar, Optional, Self
 
-from httpx import HTTPError, HTTPStatusError, RequestError
+from httpx import HTTPError, HTTPStatusError
 from pydantic import BaseModel, validator
 
 from safir.datetime import current_datetime, format_datetime_for_logging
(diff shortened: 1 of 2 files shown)


src/safir/slack/blockkit.py
```diff
@@ -338,14 +338,14 @@ def from_exception(
             )
         else:
             message = f"{type(exc).__name__}: {str(exc)}"
-            if isinstance(exc, RequestError):
+            try:
                 return cls(
                     message,
                     method=exc.request.method,
                     url=str(exc.request.url),
```
jonathansick, line 345: Are we using exception handling here instead of a `hasattr` test because sometimes there is a `request` attribute but it's `None`?

The main thing I might recommend is just adding a comment that not all RequestError exceptions have a `request`.

jonathansick (approved):

src/safir/slack/blockkit.py
rra: Yup, exactly. What happens is that it gets set to `None` in the constructor and then later exception handling adds it via attribute assignment, but that attribute assignment doesn't always happen. I'll add a comment explaining that.

rra: And hm, maybe I should just explicitly check for `None`? Using an exception handler is more comprehensive, but kind of confusing.

rra: Oh, right, now I remember: you can't check for `None` because the actual data is hidden, and `request` is a property that raises `RuntimeError` if the request hasn't been set. I'll add that to the comment as well.

PR #168, opened 2023-05-16
DM-39232: Add support for creating PVCs to Kubernetes mock

rra:
Add create_namespaced_persistent_volume_claim to the Kubernetes mock.


athornton (approved):
LGTM.  If we need event support we can add it later, but currently nothing actually waits on individual volumes to be created.

PR #169, opened 2023-05-17
DM-39251: Switch to scriv for change log management

rra:
scriv maintains change logs in individual fragments that can be merged with the corresponding PR without creating merge conflicts, and then assembles those fragments into a new change log entry during the release process. Move the pending changes to a scriv fragment, add a scriv configuration, and update the development and release documentation to document use of scriv.

Also note in the preamble of the change log that Safir follows semver.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Change log
 
-<!--
-Headline template:
-X.Y.Z (YYYY-MM-DD)
--->
+All notable changes to Safir will be documented in this file.
+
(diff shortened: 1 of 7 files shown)


rra:
The doc build is failing because the changelog.d directory doesn't exist yet.

jonathansick (approved):
This looks good, thanks.

Separately, do you think we'd want to render the "next" change log entries in the docs? That is, have the Sphinx build (for a non-release in GitHub Actions) automatically run a `scriv collect`. I guess the challenge here is computing a next version and noting that they're unreleased changes.

PR #176, opened 2023-05-18
DM-39249: Support standard Kubernetes delete options

rra:
Add support for the keyword parameter grace_period_seconds and a body of V1DeleteOptions for all delete_* methods in the Kubernetes API mock.

Stop requiring propagation_policy be set for delete_namespaced_job, and require it be given as a keyword parameter.


athornton (approved):
LGTM; we might just want to bite the bullet on the propagation policy and support all three values, but treating "Foreground" and "Background" as the same.  I don't know whether real life for unspecified propagation policy is "Foreground" or "Background" but it's definitely not "Orphan".

rra:
Yup, good idea. Done.

PR #179, opened 2023-06-15
DM-39627: Fix GithHubAppClientFactory parameter names in docs

rra:
The documentation used different parameter names for the GitHubAppClientFactory constructor than the code. Make the documentation match the code.


jonathansick (approved):
Aha, thanks!

PR #183, opened 2023-07-17
Pin major versions of dependencies

rra:
Add pins on the major version of each dependency. For those with versions less than 1, pin on <1 even though semver says the minor versions may have backward-incompatible changes, since those dependencies (FastAPI, Starlette, httpx) mostly don't do that and can have a lot of version churn.

Also pin documenteer <1, and move the lower version bound to a non-beta release because referencing any alpha or beta release in a dependency tells pip it's allowed to install any other alpha or beta release instead of sticking only to full releases.

Move the lower bound for the Redis dependency past the version that fixes typing to simplify the version constraint. There doesn't seem to be a good reason to use old versions of the Redis library.


jonathansick (approved):
Excellent; I look forward to seeing how this pans out.

PR #190, opened 2023-08-30
DM-40567: Add a helper function for Click command help

rra:
For the Phalanx command-line tool, I added support for a second level of commands to the standard help boilerplate. The result is complex enough that it seems worth lifting into Safir proper rather than existing only in a template. The result is a new safir.click module providing a display_help function.

Add documentation of the run_with_asyncio decorator to the new user guide page about Click support, since this is the most common place where that decorator is used.


rra:
The apparently-spurious explicit declaration of the return value works around a Sphinx bug that seems to be inserting the return value rST in the wrong place when an Examples section exists. This is about the third time I've seen this bug in various places; they seem to keep fixing it and then reintroducing it.

jonathansick (approved):
Very nice, thanks.

PR #194, opened 2023-09-05
DM-40628: Switch to Ruff for linting

rra:
Convert to Ruff for linting instead of isort and flake8, and fix the problems it found that seem legitimate. Add a couple of Safir-specific ignores for exception naming deviations and use of global variables in the logging setup.

Refactor the list methods in the Kubernetes mock to avoid Ruff complexity warnings and share more code. There is one remaining complexity warning that will be fixed in a follow-up PR.

Add pre-commit checks for trailing whitespace and merge conflicts, and fix some trailing whitespace.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,0 @@
-[flake8]
-max-line-length = 79
-# E203: whitespace before :, flake8 disagrees with PEP-8
-# W503: line break after binary operator, flake8 disagrees with PEP-8
-ignore = E203, W503
-exclude =
-    docs/conf.py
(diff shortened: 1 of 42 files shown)


rra:
This is mostly mechanical and uninteresting to review, but I wanted to give you a chance to object to the rewording of docstrings to make the first line fit on one line. Our documentation coding style allows multiple lines, but the default Ruff configuration doesn't. I looked at the things it flagged and it seemed straightforward to reword them without much loss, so I took that route, but I don't have strong feelings about it.

rra:
Per discussion, excluded the Ruff warning about multiline docstring summary lines and reverted the changes to satisfy that.

jonathansick (approved):

jonathansick:
The gist of our decision on docstrings is that the parsers are capable of interpreting a sentence of multiple lines as the docstring summary line, so there isn't much practical advantage in adhering to PEP 257/numpydoc's "single line" rule if it means we have to compromise on clarity and formatting to meet a character requirement on that sentence.

PR #195, opened 2023-09-05
DM-40638: Add AsyncMultiQueue data structure

rra:
Add an asyncio multiple writer, multiple reader queue, and convert the Kubernetes mock to use it instead of implementing its own. This data structure is also used in the JupyterHub REST spawner and in the Nublado lab controller.

### DIFF
--- a/changelog.d/20230905_090034_rra_DM_40638.md
+++ b/changelog.d/20230905_090034_rra_DM_40638.md
@@ -1,0 +1,3 @@
+### New features
+
+- Add new `safir.asyncio.AsyncMultiQueue` data structure, which is an asyncio multi-reader queue that delivers all messages to each reader independently.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10,6 +10,7 @@ API reference
 
 .. automodapi:: safir.asyncio
    :include-all-objects:
+   :inherited-members:
 
 .. automodapi:: safir.click
    :include-all-objects:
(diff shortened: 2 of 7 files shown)


rra:
Sphinx plus our extensions is refusing to generate any method documentation for `AsyncMultiQueue` and I cannot figure out why. If you have a few minutes, I'd love another pair of eyes on it to try to figure out what I'm doing wrong or what weird edge case I'm running into.

jonathansick:
I'm not sure about the docs build. I tried simplifying the class; I tried upgrading to Sphinx 7... so far nothing seems to be working.

jonathansick:
Dropping `sphinx_autodoc_typehints` didn't help either.

rra:
It *feels* like an `automodsumm` problem, but I can't figure out what would be special about this class to make it think it shouldn't generate method documentation.

jonathansick:
It seems that the `Generic` superclass is the issue. It's interesting that the `PydanticRedisStorage` class works fine, but it's generic has a bound?

rra:
I was wondering about that, but that was exactly my confusion. I added a bound of `BaseModel` locally and rebuilt the docs and still, no method documentation at all, so I'm not sure what's different between the two classes.

rra:
Ah, it is a bug but I found a workaround. I have to flag it with `:inherited-members:` or it tries to use `__slots__` to discover the class members and `Generic` also uses slots in some weird way that apparently sometimes confuses it.

jonathansick (approved):
Very neat class.

PR #204, opened 2023-09-12
DM-40744: Convert to Pydantic v2

rra:
Update all Pydantic code and examples to Pydantic v2. Use `validation_alias` instead of `env` for configuring the environment variables in examples that use `BaseSettings`. Convert `validate_exactly_one_of` to a model validator and require that it be called with `mode="after"`.

Pydantic now distinguishes between `str` and `int`, and the GitHub IDs all appear to be `int`s, so change the GitHub App and webhook models to type those fields as `int` rather than `str`.

### DIFF
--- a/tests/github/webhooks_test.py
+++ b/tests/github/webhooks_test.py
@@ -4,11 +4,21 @@ from __future__ import annotations
 
 from pathlib import Path
 
-from safir.github import webhooks
 from safir.github.models import (
     GitHubCheckRunStatus,
     GitHubCheckSuiteConclusion,
     GitHubCheckSuiteStatus,
+)
+from safir.github.webhooks import (
+    GitHubAppInstallationEventAction,
+    GitHubAppInstallationEventModel,
+    GitHubAppInstallationRepositoriesEventModel,
+    GitHubCheckRunEventAction,
+    GitHubCheckRunEventModel,
+    GitHubCheckSuiteEventModel,
+    GitHubPullRequestEventAction,
+    GitHubPullRequestEventModel,
+    GitHubPushEventModel,
 )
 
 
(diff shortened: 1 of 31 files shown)


rra:
The GitHub id typing change may be incorrect. I looked at the GitHub API documentation, but they don't have detailed type information for the hook messages that I can find. I *think* it's always an int and this typing is correct, but it's possible that we should do something more complicated.

tests/github/webhooks_test.py
```diff
@@ -4,12 +4,22 @@
 
 from pathlib import Path
 
-from safir.github import webhooks
 from safir.github.models import (
     GitHubCheckRunStatus,
     GitHubCheckSuiteConclusion,
     GitHubCheckSuiteStatus,
 )
+from safir.github.webhooks import (
```
rra, line 12: Sorry about this; it was just because `GitHubAppInstallationRepositoriesEventModel` is otherwise long enough that one of the code lines was unavoidably over 80 characters.

jonathansick (approved):
Looks great, thank you. The new Pydantic API will take a bit of getting used to. Definite improvements and also a bit more verbosity in places.

For the `id`s in GitHub models, and especially the app installation ID, I think I made those `str` types because Gidgethub's `gidgethub.apps.get_jwt` treats the ID as a str. Hence `GitHubAppClientFactory` also treats the ID as a str. I wonder if that factory then should treat the ID as an `int` for consistency with the models and transform that internally into a `str` for Gidgethub?

rra:
The plan for the GitHub client factory sounds good to me. I have implemented that and added a note to the class and a changelog entry, just in case that causes problems.

PR #212, opened 2023-10-03
DM-23878: Allow log level enum with any case

rra:
When configuring a FastAPI application with debug logging, I used "debug" instead of "DEBUG" in the configuration and was then annoyed that the server didn't start with an error. The function call supported a string with any case, but the Pydantic model we use to validate most configurations required uppercase because it relied on the enum to do the validation.

Implement a `_missing_` method for the LogLevel enum that converts strings in any case to the expected uppercase, which allows Pydantic validation of models to work with any case of the log level and still validates and returns the correct log level.


jonathansick (approved):
Neat, I didn't know about `_missing_`. I guess another way would be to do a pre validator in Pydantic, but this is more portable. 👍

PR #217, opened 2023-11-10
DM-41630: Convert to FastAPI lifespan functions

rra:
As of FastAPI 0.93.0, FastAPI supports a lifespan async context manager to do startup and shutdown. The part before the yield is run during application startup and the part after the yield is run during application shutdown. The older on_event annotations are now deprecated.

Add a development dependency on FastAPI 0.93.0 and change the test suite uses of on_event annotations to lifespan functions. Update all of the documentation to use the lifespan function instead of on_event handlers.

This change does not bump the Safir FastAPI dependency because Safir's own code still works fine with either method.

### DIFF
--- a/docs/user-guide/arq.rst
+++ b/docs/user-guide/arq.rst
@@ -19,21 +19,26 @@ Quick start
 Dependency set up and configuration
 -----------------------------------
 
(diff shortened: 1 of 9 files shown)


jonathansick (approved):
Great, thank you for keeping this up to date.

PR #219, opened 2023-11-15
DM-41708: Add Python 3.12 support

rra:
Add Python 3.12 to the test matrix and advertise that it is supported.


rra:
arq doesn't have pre-built wheels for Python 3.12 yet, so holding off on this until that's the case.

rra:
Holding off on this until more dependencies are ready.

PR #230, opened 2024-01-17
DM-42527: Rewrite middleware as pure ASGI middleware

rra:
Rewrite CaseInsensitiveQueryMiddleware and XForwardedMiddleware as pure ASGI middleware rather than using the Starlette BaseHTTPMiddleware class. The latter seems to be behind some poor error reporting of application exceptions, has caused problems in the past due to its complexity, and is not used internally by Starlette middleware.


jonathansick (approved):
Neat. Nice that it doesn't look too hard to write directly against ASGI here.

PR #235, opened 2024-02-01
DM-42714: Use Annotated for dependency arguments

rra:
FastAPI now recommends using Annotated to indicate arguments that are set by FastAPI via dependency injection. Among other benefits, this aligns Python's opinion about whether the argument has a default with reality when the function is called outside of a FastAPI context, which can avoid bugs in some edge case situations. It also allows removing special Ruff configuration whitelisting the special FastAPI dependency functions, since Annotated makes the intent obvious.

Also make the same change to route definitions in the test suite.

Adjust the docstring for auth_delegated_token_dependency to use a proper link with anchor text.

### DIFF
--- a/docs/user-guide/arq.rst
+++ b/docs/user-guide/arq.rst
@@ -180,7 +180,7 @@ The `safir.dependencies.arq.arq_dependency` dependency provides your FastAPI end
 
 .. code-block:: python
 
-    from typing import Any
+    from typing import Annotated, Any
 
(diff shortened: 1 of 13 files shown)


jonathansick (approved):
Looks great. My only thought here is that there are some reStructuredText pages that have examples that don't use Annotated. We could clean those up later, of course.

rra:
Good catch, thanks! I forgot about those. I think I caught them all.

PR #242, opened 2024-03-13
DM-43288: Ensure per-request database sessions are closed

rra:
In the db_session_dependency FastAPI dependency, do database session cleanup in a finally block. This will hopefully ensure that the sessions are cleaned up even on uncaught exceptions, which in turn fixes warnings in test cases in other packages.


jonathansick (approved):
Good catch!

PR #246, opened 2024-05-20
DM-44444: Allow database passwords to be SecretStrs

rra:
When initializing a database engine with create_database_engine or create_sync_session, allow the database password to be a SecretStr instead of a str. This simplifies the call for applications that use Pydantic with SecretStr for secrets.

Fix some places where development dependencies were out of date, and pin requests since the latest version breaks the docker library.


jonathansick (approved):
This all looks good, thanks.

PR #247, opened 2024-05-21
DM-44444: Add parse_timedelta function to parse durations

rra:
Add a new safir.datetime.parse_timedelta function that parses a human-friendly syntax for time durations into a datetime.timedelta. The syntax was taken from semaphore and is now also used in the configuration parser for Gafaelfawr.

### DIFF
--- a/docs/user-guide/datetime.rst
+++ b/docs/user-guide/datetime.rst
@@ -75,3 +75,41 @@ Safir therefore also provides `safir.datetime.format_datetime_for_logging`, whic
 
 As the name of the function indicates, this function should only be used when formatting dates for logging and other human display.
 Dates that may need to be parsed again by another program should use `~safir.datetime.isodatetime` instead.
+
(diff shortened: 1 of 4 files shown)


docs/user-guide/datetime.rst
```diff
@@ -75,3 +75,41 @@ Safir therefore also provides `safir.datetime.format_datetime_for_logging`, whic
...
+
+
+   @field_validator("lifetime", mode="before")
+   @classmethod
+   def _validate_lifetime(
+       cls, v: str | float | timedelta
+   ) -> float | timedelta:
+       if not isinstance(v, str):
+           return v
+       return parse_timedelta(v)
```
jonathansick, line 113: To make the example easier to follow, maybe add an example Pydantic model itself with the `lifetime` attribute to show how the field validator gets hooked up.

jonathansick (approved):
Thanks!

PR #249, opened 2024-05-30
DM-44444: Remove version cap on structlog

rra:
structlog [uses CalVer](https://github.com/hynek/structlog/blob/main/.github/SECURITY.md) and changes the major version for each year. The structlog developers recommend letting the version float, since the major version has no semantic significance. Backwards compatibility is handled through deprecation warnings.


rra:
Tests should pass once rebased on top of https://github.com/lsst-sqre/safir/pull/248.

PR #250, opened 2024-05-30
DM-44444: Use Python 3.12 for GitHub actions

rra:
For all GitHub Actions that do not use a version matrix to ensure that the code itself works with older versions, use Python 3.12 as the Python version, under the principle that tests unrelated to compatibility should follow the latest supported Python version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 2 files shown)


rra:
Tests should pass once rebased on top of https://github.com/lsst-sqre/safir/pull/248.

jonathansick (approved):
Good idea. A while ago I read about GitHub's [Variables](https://docs.github.com/en/actions/learn-github-actions/variables) feature. That could be a nice way to dry out variables like the default Python version. Works either on a per workflow and a repo-wise level I believe. Sorry I haven't taken the initiative to PR this myself already!

rra:
Oh, that looks interesting. I'll go ahead and merge this for now, since that looks a bit more complicated to set up, but will then look into it more.

PR #261, opened 2024-07-05
DM-45137: Update GitHub Actions settings from the template

rra:
Use an environment variable for the default Python version with which to run most actions. Add timeouts to jobs that were missing them. Use the tox-uv plugin uniformly. Fix one place where merge queues weren't handled correctly. Update the periodic CI notification configuration based on proposed template updates.


jonathansick (approved):
Great, thanks.

PR #262, opened 2024-07-05
DM-45137: Update dependencies, switch to uv

rra:
Update pre-commit dependencies.

Switch to uv for dependency installation in the Makefile, and add update and update-deps targets for parallelism with FastAPI application packages, even though update-deps doesn't do much (just update uv and pre-commit and run pre-commit autoupdate).


jonathansick (approved):
This is great thanks. Does the `run-tox` action in the GHA workflows also need an update to make sure the `tox-uv` plugin is installed there?

jonathansick:
Nevermind, I see it's covered in #261

PR #263, opened 2024-07-05
DM-45137: Switch to a shared Ruff configuration

rra:
Copy ruff-shared.toml from the new version of the project template and extend it with Safir-specific rules. This will hopefully make it easier to update the shared configuration in multiple repositories.


jonathansick (approved):
This is great. Someday we can rig up an updater either as a squarebot-driven PR, a reusable workflow that checks for the shared rules being up to date, or a more basic make/tox/nox driven command that pulls an updated shared configured file.

PR #265, opened 2024-07-10
DM-45138: Allow custom serializers for Redis arq queue

rra:
For UWS applications, to reduce pickle problems between different versions of libraries and Python, we plan to use a custom serializer that uses Pydantic's JSON serialization. arq supports custom serializers, but they weren't supported by RedisArqQueue. Plumb those arguments through the initialize class method.


rra:
Found a different way of doing this that is contained within vo-cutouts and has considerably fewer issues.

PR #266, opened 2024-07-11
DM-45138: Add support for aborting arq jobs

rra:
Add a new `abort_job` method to instances of `safir.arq.ArqQueue`, which tries to abort a job. If the job has already started, this requires the worker to enable support for aborting jobs.

The mock implementation deletes the job entirely if it hasn't been started, and treats it as if it failed with asyncio.CancelledError if it were already running. This seems to match what arq does with a Redis queue.


jonathansick (approved):
Great new feature, thanks.

PR #269, opened 2024-07-16
DM-45281: Add new timedelta data types for Pydantic models

rra:
Take advantage of the new functional validator support in Pydantic v2 and encapsulate custom `timedelta` validation as reusable types. The `SecondsTimedelta` type accept integer or floating-point number of seconds, possibly stringified. The `HumanTimedelta` type accepts the same, plus the duration strings accepted by `parse_timedelta`.

Unfortunately, sphinx-automodapi apparently doesn't recognize `TypeAlias` objects as something for which documentation should be generated, even with `:include-all-objects:`, so references to these new types have to be ignored when generating documentation.


jonathansick (approved):
The pydantic functional validators do seem like a natural solution and a replacement for the custom types that I was sometimes implementing previously in Pydantic 1. Its too bad that they can't be documented automatically.

PR #270, opened 2024-07-18
DM-45281: Add Pydantic types for Postgres and Redis DSNs

rra:
Add new types `safir.pydantic.EnvAsyncPostgresDsn` and `safir.pydantic.EnvRedisDsn`, which validate PostgreSQL and Redis DSNs but rewrite them based on the environment variables set by tox-docker. Programs using these types for their configuration will therefore automatically honor tox-docker environment variables when running the test suite. `EnvAsyncPostgresDsn` also enforces that the scheme of the DSN is compatible with asyncpg and the Safir database support.


jonathansick (approved):
Very nice! I like this solution to the tox-docker environment variable issue.

PR #271, opened 2024-07-18
DM-45281: Break apart `safir.database` for ease of maintenance

rra:
Convert `safir.database` from a single-file module to a directory and separate the code into internal modules for ease of maintenance. Upcoming PRs will be adding more code to `safir.database`, and this will avoid dumping all of that code into a single large file.

### DIFF
--- a/src/safir/database/__init__.py
+++ b/src/safir/database/__init__.py
@@ -1,0 +1,16 @@
+"""Utility functions for database management."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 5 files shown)


rra:
This is purely moving code around, so review is just conceptual. If you think this is a good idea, I can put up a PR to break apart some of the other larger modules.

jonathansick (approved):
Sounds good. I like this approach for growing modules.

PR #272, opened 2024-07-18
DM-45281: Add decorator to retry async transactions

rra:
Add `safir.database.retry_async_transaction`, which retries a function or method a configurable number of times if it fails with a SQLAlchemy exception indicating a problem with the underlying database layer, such as that raised when there is a conflicting transaction.

### DIFF
--- a/docs/user-guide/database.rst
+++ b/docs/user-guide/database.rst
@@ -433,9 +433,38 @@ Setting an isolation level
 
 `~safir.database.create_database_engine` and the ``initialize`` method of `~safir.dependencies.db_session.db_session_dependency` take an optional ``isolation_level`` argument that can be used to set a non-default isolation level.
 If given, this parameter is passed through to the underlying SQLAlchemy engine.
-See `the SQLAlchemy isolation level documentation <https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#setting-transaction-isolation-levels-dbapi-autocommit>`__ for more information.
+See `the SQLAlchemy isolation level documentation <https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#setting-transaction-isolation-levels-dbapi-autocommit>`__ for more information.
 
 You may have to set a custom isolation level, such as ``REPEATABLE READ``, if you have multiple simultaneous database writers and need to coordinate their writes to ensure consistent results.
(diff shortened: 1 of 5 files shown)


src/safir/database/_retry.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    def decorator(f: RetryF) -> RetryF:
+        @wraps(f)
+        async def wrapper(
+            *args: RetryP.args, **kwargs: RetryP.kwargs
+        ) -> RetryT:
+            for _ in range(1, max_tries):
+                with contextlib.suppress(DBAPIError):
+                    return await f(*args, **kwargs)
```
jonathansick, line 95: There isn't any sort of sleep/delay between tries. Is the assumption that the event loop provides sufficient waiting? What if there are multiple pods and some of them aren't active so their event loop is very fast? I might be missing the concept here.

jonathansick (approved):

src/safir/database/_retry.py
rra: Oh, that's an interesting question. My mental model was that if two modifications happened at the same time, they would both be able to open transactions and then whichever one finished first would win and the other one would get an exception raised when it tried to commit, but at that point the other transaction has gone through so an immediate retry is fine. But maybe that's not how the underlying database model actually works. I guess I don't actually know.

docs/user-guide/database.rst
```diff
@@ -433,9 +433,38 @@ Setting an isolation level
...
+
+   from safir.database import datetime_to_db, retry_async_transaction
+   from sqlalchemy.ext.asyncio import async_scoped_session
+
+
+   class Storage:
+       def __init__(self, session: async_scoped_session) -> None:
+           self._session = session
+
+       @retry_async_transaction
```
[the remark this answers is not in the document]
rra: Yeah, that's a good idea, I'll add that.

src/safir/database/_retry.py
jonathansick: Ah. I'm totally happy going with this because I don't have a strong notion of what would happen but your model seems quite reasonable.

rra: After thinking about this for a bit, I think you're right and it makes sense to add a short delay just on general principles. I usually do that with retries, and I think a half-second delay would be enough for things to clear and to prevent us from hammering too hard on the database with retries.

PR #277, opened 2024-07-23
DM-45281: Switch to testcontainers from tox-docker

rra:
In preparation for converting Safir to a monorepo with multiple PyPI modules and thus the nox build system, switch from tox-docker to testcontainers for testing.


jonathansick (approved):
Looks good!

PR #279, opened 2024-07-23
DM-45281: Convert to nox for the build system

rra:
In preparation for converting Safir to a monorepo with multiple PyPI modules, convert to nox for the build system and move the current code to a directory named safir. This will hold the main `safir` package.


jonathansick (approved):
Looks good!

PR #280, opened 2024-07-23
DM-45281: Separate safir-arq from safir

rra:
Move the `safir.arq` module into a separate package that can be independently distributed on PyPI. This allows the arq support library to be installed in contexts where the Safir dependency is too heavy-weight.

Convert all of the submodules of Safir into directories so that they can include `py.typed` files now that the safir package is a namespace-only package. (mypy's documentation says that `py.typed` is not supported in namespace packages.) When importing across the namespace package, avoid using relative imports just for clarity and so that `safir.arq` doesn't have to be handled separately from any other submodule. Use relative imports only within the same subpackage of Safir.

Since `safir` is now a namespace-only package, drop the top-level version information from the `safir` package. I don't think we were using this; hopefully nothing relies on it and is using the installed package metadata instead.


jonathansick (approved):
Looks great. The change to needing subpackages for everything is an interesting side-effect but seems to not be a problem.

PR #281, opened 2024-07-26
DM-45281: Add UWS support library

rra:
Add `safir.uws`, which provides a general framework for writing UWS applications with backend workers that use arq. The backend worker function can be any sync Python code and the backend image can be different from the frontend image.

Add `safir.arq.uws`, which contains only the code required by the backend worker with a minimum of dependencies so that it can be installed on backend workers where more complex dependencies from Safir may conflict.

Tested by converting vo-cutouts to this version of the library and checked that it still passed its test suite.


safir/src/safir/uws/_service.py
```diff
@@ -0,0 +1,601 @@
...
+        behavior, and waiting for completion, to use as a building block when
+        constructing a sync API.
+
+        Parameters
+        ----------
+        user
+            User on behalf this operation is performed.
+        job_id
+            Identifier of the job.
+        wait
```
stvoutsin, line 221:
```suggestion
        wait_seconds
```

safir/src/safir/uws/_storage.py
```diff
@@ -0,0 +1,424 @@
...
+    async def mark_failed(
+        self, job_id: str, exc: Exception, *, end_time: datetime | None = None
+    ) -> None:
+        """Mark a job as failed with an error.
+
+        Parameters
+        ----------
+        job_id
+            Identifier of the job.
+        error
```
stvoutsin, line 322:
```suggestion
        exc
```

safir/src/safir/uws/_service.py
```diff
@@ -0,0 +1,601 @@
...
+            logger.warning("Job timed out", timeout=self._config.sync_timeout)
+            raise SyncJobTimeoutError(self._config.sync_timeout)
+        if job.error:
+            logger.warning("Job failed", error=job.error.to_dict())
+            raise SyncJobFailedError(job.error)
+        if not job.results:
+            logger.warning("Job returned no results")
+            raise SyncJobNoResultsError
+
+        # Return the the first result.
```
stvoutsin, line 365:
```suggestion
        # Return the first result.
```

safir/src/safir/uws/_service.py
```diff
@@ -0,0 +1,601 @@
...
+            some subclass of `pydantic.BaseModel`.
+
+        Raises
+        ------
+        safir.uws.UWSError
+            Raised if there is some problem with the job parameters.
+        """
+        return self._config.parameters_type.from_job_parameters(params)
+
+    async def _wait_for_job(
```
stvoutsin, line 567: While this is perfectly fine, would it make sense to add a max delay to make sure it doesn't grow too large? Also perhaps using async.timeout here?
I haven't tested this, but here's the idea:

```
async def _wait_for_job(
    self, job: UWSJob, until_not: set[ExecutionPhase], timeout: timedelta
) -> UWSJob:
    async def check_job():
        nonlocal job
        delay = 0.1
        max_delay = 5.0 
        increase_factor = 1.5
        
        while job.phase in until_not:
            await asyncio.sleep(delay)
            job = await self._storage.get(job.job_id)
            delay = min(delay * increase_factor, max_delay)

    with suppress(asyncio.TimeoutError):
        await asyncio.wait_for(check_job(), timeout.total_seconds())

    return job

```

stvoutsin (changes requested):
Looks good to me, only minor suggestions (mostly typos in docstrings)
Overall it all makes sense, and at first glance I can't spot anything wrong or invalid based on the UWS spec

safir/src/safir/uws/_service.py
rra: This is a great point, thank you! I don't know why I didn't think to use `asyncio.timeout` for this. I've adopted a variation of this approach; it's a lot more compact and straightforward.

safir/src/safir/uws/_exceptions.py
```diff
@@ -0,0 +1,277 @@
...
+            case WorkerErrorType.FATAL:
+                error_code = ErrorCode.ERROR
+                error_type = ErrorType.FATAL
+            case WorkerErrorType.TRANSIENT:
+                error_code = ErrorCode.SERVICE_UNAVAILABLE
+                error_type = ErrorType.TRANSIENT
+            case WorkerErrorType.USAGE:
+                error_code = ErrorCode.USAGE_ERROR
+                error_type = ErrorType.FATAL
+                slack_ignore = True
```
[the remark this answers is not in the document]
rra: Thankfully, mypy will check that a match/case statement is exhaustive, so it's best to not include the default case if the other cases are supposed to be exhaustive. That ensures mypy will raise an error if another possibility arises in the future.

rra:
Thanks for the review! I've also committed all the changes to docstrings and the comment fix.

safir/src/safir/uws/_exceptions.py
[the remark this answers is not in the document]
stvoutsin: Ah ok good point

stvoutsin (approved):
Looks good to me!

PR #288, opened 2024-08-16
DM-45824: Split safir.logging into a separate PyPI package

rra:
In preparation for adding logging configuration to UWS arq workers, which should not depend on the full `safir` package, split `safir.logging` into a new PyPI package `safir-logging`. `safir` depends on this package, so this should not affect backward compatibility.

This requires some additional complexity in the integration between logging configuration and the logger dependency, which is done via a conditional import and skipped if the `safir.dependencies` module is not available.

Mention in the arq documentation that `safir.arq` is also available via the `safir-arq` package, noticed when making a similar update for `safir.logging`.

Change the reference to a `Configuration` object in the logging documentation to `Config` instead, now that we've standardized on that naming for our Safir-based applications.

Drop the Safir dependency on `psycopg2`, which was left over from sync database support that was removed in 6.0.0.


jonathansick (approved):
This looks good, thanks.

I forgot that we'd changed the configuration model's name to Config in our templates from Configuration. But fair enough.

PR #294, opened 2024-08-30
DM-46034: Reorganize database documentation

rra:
Break up the long database documentation page into pages focused on specific topics. Get rid of the arbitrary heading for "less-used operations" as part of that restructuring.

Reorganize the documentation about retrying database transactions to hopefully be a bit clearer.


rra:
This is almost all just moving information around, so the only things to review here are the new structure and the tweaks to the way transaction retrying is described on the `retry.rst` page. Documentation for the testcontainers method of database testing is coming in a subsequent PR.

This PR is in preparation for adding documentation on how to use Alembic.

jonathansick (approved):
Looks like a good re-org to start with.

PR #297, opened 2024-09-10
DM-46034: Reorganize UWS database schema

rra:
Using a directory named `_schema` with private modules inside it does not work correctly with Sphinx and automodapi. Sphinx is unable to find the imported and re-exported symbols from those private modules.

The separation of the schema into separate files is unnecessary, so merge them all together into a single `_schema.py` module file, which does work correctly with Sphinx.

Convert to the new style of explicit inheritance from DeclarativeBase and use `Mapped`, `mapped_column`, and type derivation from type annotations instead of `Column`. This is the new, recommended way of doing ORM definitions in SQLAlchemy. Bump the version dependency for `safir[db]` to SQLAlchemy 2.0.0 or later since some of the elements of this new style were only added in the runup to the 2.0.0 release.


rra:
This is in preparation for adding Alembic support, which required exposing the declarative base of the schema as an exported symbol. The old-style declarative base didn't work well with Sphinx, and that seemed like a good opportunity to redo the schema using the new, recommended syntax.

jonathansick (approved):
Looks good. I'll have to to look into adopting the new table declaration syntax.

I really have no idea why autodoc/automodapi aren't finding types in a subpackage named `_schema`. One day I'd love to have the time to track down all these things.

rra:
There's some special code... argh, somewhere, I forget where I found it now, I think in automodapi? that recognizes the pattern of a symbol imported from a module starting with an underscore and artificially removes one level of imports when building the canonical name for documentation cross-reference purposes, and I think it explicitly only handles doing that once and not twice.

jonathansick:
oh dear. That does sound like a brittle approach.

PR #299, opened 2024-09-11
DM-46034: Add support functions for Alembic

rra:
Add functions that support use of Alembic in Safir-based applications. These include utility functions for the Alembic `env.py` file that work with Safir's database support, functions to mark a database as current for Alembic when initializing it, and a function to check whether the database schema is up-to-date that can be used during application startup.

Add `configure_alembic_logging` to safir.logging, which configures Alembic to use structlog and JSON for logging, aiding parsing of the logs by Google's Cloud Logging system.

Add optional support for Alembic to the UWS library. This is not yet documented in the manual; that will be done separately as part of integrating support into the Safir FastAPI app template.


docs/user-guide/database/schema.rst
```diff
@@ -0,0 +1,465 @@
+################
+Managing schemas
+################
```
jonathansick, line 3:
```suggestion
##########################
Managing schema migrations
##########################
```

I think *migrations* is the key word for this page.

jonathansick (approved):
Looks great. Thanks for making Alembic support easy to add.

PR #300, opened 2024-09-12
DM-46034: Remove remaining declarative_base references

rra:
Update the documentation for defining a database schema and one test case to use the new `DeclarativeBase` inheritance syntax. Add a documentation link to the current way of defining ORM models with SQLAlchemy.


rra:
There's one remaining use of `declaractive_base` that will be removed by https://github.com/lsst-sqre/safir/pull/299.

jonathansick (approved):

PR #302, opened 2024-09-19
DM-45522: Kafka connection helpers

fajpunk:
For metrics events publishing, we've decided that we want to connect directly to Kafka. Here are some helpers to do that from Safir apps.

The Sasquatch/Phalanx docs that go with this are coming soon...

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,4 @@ Read more about Safir at https://safir.lsst.io.
 
 ## Developing Safir
 
-The best way to start contributing to Safir is by cloning this repository creating a virtual environment, and running the init command:
-
-```sh
-git clone https://github.com/lsst-sqre/safir.git
-cd safir
-make init
-```
-
-For details, see https://safir.lsst.io/dev/development.html.
+See https://safir.lsst.io/dev/development.html.
--- a/changelog.d/20240919_141153_danfuchs_kafka.md
+++ b/changelog.d/20240919_141153_danfuchs_kafka.md
@@ -1,0 +1,3 @@
+### New features
+
+- Add new `safir.kafka module to simplify configuring and constructing Kafka clients.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -54,6 +54,9 @@ API reference
 .. automodapi:: safir.github.webhooks
    :include-all-objects:
 
+.. automodapi:: safir.kafka
+   :include-all-objects:
+
 .. automodapi:: safir.kubernetes
    :include-all-objects:
 
--- a/docs/documenteer.toml
+++ b/docs/documenteer.toml
@@ -9,12 +9,15 @@ package = "safir"
 rst_epilog_file = "_rst_epilog.rst"
 nitpicky = true
 nitpick_ignore_regex = [
+    ['py:.*', 'faststream.*'],
     ['py:.*', 'fastapi.*'],
     ['py:.*', 'kubernetes_asyncio.*'],
     ['py:.*', 'httpx.*'],
     ['py:.*', 'pydantic.*'],
     ['py:.*', 'respx.*'],
     ['py:.*', 'starlette.*'],
+    # Bug in autodoc_pydantic re: models with validators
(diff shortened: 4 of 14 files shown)


jonathansick (commented):
This is a conceptual design comment, so let me know if you find the idea interesting or not before we get into the specifics:

I'm wondering about the new KafkaConnectionSettings class and the `validated` property strategy. I see where you're going with this, but the drawback is duplicating fields across new models. This also kind of tightly connects the KafkaConnectionSettings usage into the factory functions because of the need to deal with the different types.

The factory functions are also tricky for a couple reasons. First, they don't (but maybe should?) allow for extra keyword arguments to the constructors. Second, the Squarebot apps will create a faststream fastapi KafkaRouter instead of a KafkaBroker.

I'd be slightly more in favor of enhancing the KakfkaConnectionSettings class itself so that we're less reliant on the factory functions. So the KafkaConnectionSettings would have an after validator that checks the different settings combinations to ensure that each auth method has the right settings set. And there could be a property on KafkaConnectionSettings that returns an enum with an authentication strategy type.

Another thing this settings class could do is have methods that return a dict of arguments for the various clients that set authentication keyword arguments. For faststream it can create the BaseSecurity directly. This might even obviate some of the factory functions.

fajpunk:
@jonathansick 
> I'm wondering about the new KafkaConnectionSettings class and the validated property strategy. I see where you're going with this, but the drawback is duplicating fields across new models
> So the KafkaConnectionSettings would have an after validator that checks the different settings combinations to ensure that each auth method has the right settings set.
In that case (if I'm understanding what you're saying), you still have to enumerate the valid property combinations, they're just in conditional statements.  It's not any more repetitive to enumverate them in Pydantic models, we get nice error messages for which (maybe multiple) attributes are missing.

> Another thing this settings class could do is have methods that return a dict of arguments for the various clients that set authentication keyword arguments. For faststream it can create the BaseSecurity directly. This might even obviate some of the factory functions.
This is a great point, and I agree that this would obviate the factory functions.  My intention with the factory functions is that you get a really easy way of getting a happy path client, but getting dict args helps just as much in client construction, and is much more flexible.

jonathansick:
> In that case (if I'm understanding what you're saying), you still have to enumerate the valid property combinations, they're just in conditional statements. It's not any more repetitive to enumerate them in Pydantic models, we get nice error messages for which (maybe multiple) attributes are missing.

Right, I was originally thinking that the validator would itself go through each auth type and see if the settings were available for it, and you'd have to write a custom ValueError to report if nothing was possible. So I see how this sub-model approach lets us just use Pydantic validation directory for all cases.

If we do go with a strategy where the `KafkaConnectionSettings` produces dicts with factory arguments, then that `validated` property could almost become an internal detail to produce the settings arguments, but as a user I'd only use the KafkaConnectionSettings APIs.

Specifically, it'd be neat if I could do something like this to create a KafkaRouter:

```
config = KafkaConnectionSettings()
router = KafkaRouter(config.bootstrap_servers, **config.make_faststream_router_args())
```

Or for a KafkaBroker:

```
router = KafkaRouter(config.bootstrap_servers, **config.make_faststream_broker_args())
```

Or for an AIOKafkaConsumer:

```
consumer = AIOKafkaConsumer(
    'my_topic',
    'my_other_topic',
    **config.make_aiokafka_consumer_args()
)
```

Or for an AIOKafkaProducer:

```
producer = AIOKafkaConsumer(**config.make_aiokafka_producer_args())
```

jonathansick:
Actually, for faststream the KafkaConnectionSettings could just produce a BaseSecurity instance, and that'd be enough for the faststream use case, like

```
router = KafkaRouter(config.bootstrap_servers, security=config.faststream_security)
```

fajpunk:
Yep, exactly.  I'll change this to use dict-returning methods instead.

jonathansick (approved):
Looks great! Thanks for indulging the change of approach to setting up clients, but I do think it'll work better in the long run for my purposes.

PR #308, opened 2024-10-14
DM-46821: Allow camel-case for Kafka and events configuration

rra:
Allow Kafka and event metrics configuration to be loaded from YAML files generated from Helm value files, which use camel-case instead of Python's preferred snake-case. Unforunately, using an alias generator plus an environment variable prefix doesn't work the way that one would want, nor does using an environment variable prefix with fields that have aliases, so this requires explicitly listing the camel-case and environment variable aliases for fields that support both.

### DIFF
--- a/safir/src/safir/kafka/_kafka_config.py
+++ b/safir/src/safir/kafka/_kafka_config.py
@@ -16,7 +16,14 @@ from faststream.security import (
     SASLScram256,
     SASLScram512,
 )
-from pydantic import BaseModel, Field, FilePath, SecretStr, model_validator
+from pydantic import (
+    AliasChoices,
+    BaseModel,
+    Field,
+    FilePath,
+    SecretStr,
+    model_validator,
+)
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


rra:
Gafaelfawr loads its configuration from a YAML file generated from the Helm `values.yaml`, so wants to use camel-case for these settings. This will conflict with the work you were doing on a smarter disabled config; don't worry about that, I'll rebase after you've landed your changes.

fajpunk (approved):
Config changes and cleanup lgtm.  I'm looking at the Schema metadata stuff now too...

rra:
You can ignore the schema registry stuff; I was just experimenting with a few ideas to see if it changed anything, on the basis of guesswork from the source code. Sorry to have reused this PR branch for that! I should have made a new one, but I had a Gafaelfawr build all set up with it.

safir/src/safir/metrics/_event_manager.py
```diff
@@ -227,16 +228,13 @@ async def create_publisher(
...
-                namespace = self._topic
-
         # The full, metadata-enriched event model
         event_class = create_model(
             "EventModel",
-            __base__=(payload_model, EventMetadata, MetaBase),
+            __base__=(payload_model, EventMetadata),
+        )
+        event_class._metadata = SchemaMetadata(  # noqa: SLF001
+            schema_name=name, namespace=self._topic
```
fajpunk, line 237: Here's another possible way to deal with this, without futzing with internals (though it's the same in spirit, knowing about the internals :( ):
https://github.com/lsst-sqre/safir/pull/309

PR #309, opened 2024-10-16
DM-45522 metrics: don't lose schema metadata

fajpunk:


rra (approved):
Oh, *that's* what's going on. That makes perfect sense now.

safir/src/safir/metrics/_models.py
```diff
@@ -43,7 +45,7 @@ class EventPayload(AvroBaseModel):
     """All event payloads should inherit from this."""
 
     @classmethod
-    def validate_structure(cls) -> None:
+    def validate_structure(cls: type[Any]) -> None:
```
rra, line 48: You shouldn't need this; `self` and `cls` should never get type annotations.

rra: Oh, I see the problem. Maybe just use `pydantic.create_model` here instead of the `class` syntax? I think mypy just can't handle using `cls` as the base type for a class.

fajpunk: Without it, mypy gives:

```
safir/src/safir/metrics/_models.py:78: error: Variable "cls" is not valid as a type  [valid-type]
safir/src/safir/metrics/_models.py:78: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
safir/src/safir/metrics/_models.py:78: error: Invalid base class "cls"  [misc]
Found 2 errors in 1 file (checked 154 source files)
```
https://github.com/lsst-sqre/safir/actions/runs/11357805421/job/31591269528#step:3:110

Giving it an `Any` type seems to still be the workaround:
https://github.com/python/mypy/issues/2477#issuecomment-262826099

rra: Yeah, this is fine. I may poke at it more, since something about it is bugging me, but using the `create_model` thing is just going to cause the problems with `AvroBaseModel` we saw elsewhere, so this is probably the most concise.

PR #310, opened 2024-10-16
DM-45522: Use `timedelta`s in event payloads

fajpunk:
This is admittedly somewhat nasty, and depends on a deprecated Pydantic feature (`json_encoders` in model config), but I think it's the only way to be able to serialize time deltas directly on `AvroBaseModel` models.

[This](https://github.com/marcosschroh/dataclasses-avroschema/blob/b8f7b8dfa877fea58887e4cd34d34e2ab6551afa/dataclasses_avroschema/fields/fields.py#L1000-L1038) is the code that enables this is [dataclasses-avroschema](https://github.com/marcosschroh/dataclasses-avroschema).  It requires explicitly that:
* The custom type has a `__get_pydantic_core_schema__` method
* The model has a (deprecated) `json_encoders` config set
* This `json_encoders` dict maps the custom type to a function that is itself one of the types that is [natively supported by dataclasses-avroschema](https://marcosschroh.github.io/dataclasses-avroschema/fields_specification/#avro-field-and-python-types-summary), which is why the custom type needs a `__float__` method.

This is sorta the way that Pydantic custom type serialization works, but has other requirements that are specific to `dataclasses-avroschema`, which is why it looks pretty weird (to me).

Because this is weird and depends on a deprecated Pydantic feature (that logs a warning when it's used), it may be worth trying to contribute native timedelta support to `dataclasses-avroschema`.

### DIFF
--- a/docs/user-guide/metrics/index.rst
+++ b/docs/user-guide/metrics/index.rst
@@ -110,6 +110,13 @@ We can do this all in an ``events.py`` file.
    Fields in metrics events can't be other models or other nested types like dicts, because the current event datastore (InfluxDB) does not support this.
    Basing our event payloads on `safir.metrics.EventPayload` will enable the `~safir.metrics.EventManager` to ensure at runtime when our events are registered that they don't contain incompatible fields.
 
+.. warning::
+
+   `dataclasses-avroschema`_ does not currently support timedelta fields, and will throw an exception if you define a field of type ``timedelta`` in your event payload model.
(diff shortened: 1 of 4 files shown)


rra (approved):
Ugh, that's really ugly. I don't like that dataclasses-avroschema doesn't support `Annotated` types with `Serializer` set, which is the modern way of doing this. I hate to take a deprecation warning for a somewhat optional feature. On the other hand, I think we're going to want to log `timedelta` a LOT, and this prevents people from inventing their own approaches. So I think it's probably worth it, but native support would be a lot better for sure.

PR #311, opened 2024-10-17
DM-46821: Refactor metrics configuration

rra:
Rework the metrics configuration so that `MetricsConfiguration` is a union type supporting either a disabled or a Kafka-based metrics configuration, which can be used as a drop-in type in an application `BaseSettings` class and will automatically select one or the other based on what configuration is available. Have those two configuration classes in turn create two different `EventManager` classes that provide the same interface, and two different underlying `EventPublisher` classes. Use this to implement the no-op metrics functionality rather than checking a boolean. This allows all of the Kafka configuration to be omitted if metrics are disabled.

Update the documentation accordingly, and fix a few minor problems with the documentation discovered as part of the Gafaelfawr integration.

Move the `app_name` field to the top level of the metrics configuration on the grounds that this will likely apply to any application metrics, including things that are not events. Rename `metrics_events` to just `events`, since the metrics part is clear in context.

Use the `topic_prefix` name uniformly instead of `base_topic_prefix`.


safir/src/safir/metrics/_config.py
```diff
@@ -38,45 +82,87 @@ class MetricsConfiguration(BaseSettings):
...
-        description='Set to "True" to prevent actually publishing metrics',
+    events: EventsConfiguration = Field(
+        default_factory=EventsConfiguration,
+        title="Events configuration",
     )
 
     model_config = SettingsConfigDict(
-        case_sensitive=False,
-        env_prefix="METRICS_",
+        alias_generator=to_camel,
```
fajpunk, line 91: This doesn't conflict with the [validation_alias](https://github.com/lsst-sqre/safir/pull/311/files#diff-5c497f6272460b4ab381bc8d81a2801f0dc6c6fd90f945257c889e77fac4c569R82) declared on the `app_name` field?

safir/src/safir/metrics/_config.py
```diff
@@ -86,26 +172,94 @@ def make_manager(self, logger: BoundLogger | None = None) -> EventManager:
...
+)
+"""Type to use for metrics configuration in the application config.
+
+This will resolve to one of the various valid types of metrics configuration,
+all of which support a `~safir.metrics.BaseMetricsConfiguration.make_manager`
+method to create an `~safir.metrics.EventManager`.
+"""
+
+
+def metrics_configuration_factory() -> MetricsConfiguration:
```
fajpunk, line 215: Nice! This is the other place I got stuck :)

safir/src/safir/metrics/_event_manager.py
```diff
@@ -26,120 +27,335 @@
...
         Parameters
         ----------
-        Payload
-            A subclass of EventPayload to be enriched with metadata and
-            published to Kafka.
+        name
+            Name of the event. This will be used as the name of the Avro
+            schema and record and the name of the event in the event storage
+            backend.
+        topic
```
fajpunk, line 215: You can mention that this will be the Avro namespace for the record

safir/src/safir/metrics/_event_manager.py
```diff
@@ -155,215 +371,184 @@ class MyEvent(EventPayload):
...
+    ) -> EventPublisher[P]:
+        if not self._initialized:
+            msg = "Initialize EventManager before creating event publishers"
+            raise EventManagerUnintializedError(msg)
+        if name in self._publishers:
+            raise DuplicateEventError(name)
+
+        # Construct the event model.
+        model = self.build_validated_model(name, self._topic, payload_model)
+
```
fajpunk, line 547: Is it worth putting this in the base class to avoid duplication and divergence if this logic changes?

fajpunk (approved):
This is great, thanks! A couple of minor questions

safir/src/safir/metrics/_config.py
rra: No, alias generators are ignored if `validation_alias` is explicitly defined. This area of Pydantic settings is kind of confusing.

fajpunk (approved):

safir/src/safir/metrics/_event_manager.py
rra: Yeah, this is a good call, I was being lazy. :) Working on a refactor now.

safir/src/safir/metrics/_event_manager.py
```diff
@@ -155,215 +374,176 @@ class MyEvent(EventPayload):
...
+    def __init__(
+        self,
+        application: str,
+        topic_prefix: str,
+        logger: BoundLogger | None = None,
+    ) -> None:
+        super().__init__(f"{topic_prefix}.{application}")
+        self._application = application
+        self._logger = logger or structlog.get_logger("safir.metrics")
```
fajpunk, line 531: As long as you're putting some of the common `__init__` logic in the base class, why not put this there too?

fajpunk: "Because it's more clear what's going on" is a perfectly valid answer, and maybe why my version of this was so complex

rra: Yeah, this is true -- it's because if I put it in the parent class, I have to make logger a public attribute, which feels weird because things outside of the event manager shouldn't really use it. Or I leave it as a private attribute and then subclasses have to know they can use the private attribute, so it breaks the encapsulation. I sometimes make weird choices because Python doesn't really have a convention for protected attributes that are intended for subclasses but not general use. I should get over it and just move the common code. :) I'll do that.

PR #317, opened 2024-10-22
DM-47027: Change ws_post_params_dependency to handle multiple occurrences of the same param

stvoutsin:
**Summary**

PR addresses issue where the uws_post_params dependency did not correctly read in list parameters, i.e. multiple occurrences of the same param. The fix was to change items() to multi_items() on the request.form() object.
A test has been added which sets up a test endpoint which returns the params passed in to the request, and exercises it with a request with multiple of the same param.


rra (approved):
Looks good to me, thank you!

PR #326, opened 2024-11-13
DM-47262: Fix IVOA DALI timestamp parsing, add data types

rra:
Change `parse_isodatetime` and `normalize_isodatetime` to support exactly the formats specified by the IVOA DALI standard (and fix incorrect references to UWS). This means supporting timestamps without the time component and making the trailing Z optional, but requiring seconds be present if the time portion is present. This should not interfere with the other uses, such as parsing Kubernetes timestamps.

Add new data types `UtcDatetime` and `IvoaIsoDatetime` that are now the preferred way to use the normalization functions in models. The former accepts any format Pydantic supports but coerces it to UTC and timezone-aware. The latter accepts only the formats supported by `parse_isodatetime`. Document those data types instead of the normalization functions, although the normalization functions are still available (for now).

### DIFF
--- a/safir/src/safir/pydantic/_validators.py
+++ b/safir/src/safir/pydantic/_validators.py
@@ -8,6 +8,8 @@ from typing import Any, TypeVar
 
 from pydantic import BaseModel
 
+from safir.datetime import parse_isodatetime
+
 T = TypeVar("T")
 
 __all__ = [
@@ -35,6 +37,16 @@ def normalize_datetime(v: Any) -> datetime | None:
     datetime.datetime or None
         The timezone-aware `~datetime.datetime` or `None` if the input was
         `None`.
+
+    Raises
+    ------
+    ValueError
+        Raised if the input could not be parsed as a `~datetime.datetime`.
+
+    Notes
+    -----
+    Prefer to use the `~safir.pydantic.UtcDatetime` type instead of using this
+    function as a validator.
 
     Examples
     --------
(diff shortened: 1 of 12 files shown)


safir/src/safir/pydantic/_validators.py
```diff
@@ -107,12 +130,9 @@ class Info(BaseModel):
...
     if v is None:
         return None
-    if not isinstance(v, str) or not v.endswith("Z"):
-        raise ValueError("Must be a string in YYYY-MM-DDTHH:MM[:SS]Z format")
-    try:
-        return datetime.fromisoformat(v[:-1] + "+00:00")
-    except Exception as e:
-        raise ValueError(f"Invalid date {v}: {e!s}") from e
+    if not isinstance(v, str):
+        raise ValueError("Must be a string in YYYY-MM-DD[THH:MM[:SSS]] format")
```
fajpunk, line 134:
```suggestion
        raise ValueError("Must be a string in ``YYYY-MM-DD[THH:MM:SS[.mmm]][Z]`` format")
```

safir/src/safir/pydantic/_validators.py
```diff
@@ -69,24 +81,35 @@ class Info(BaseModel):
...
     """Pydantic field validator for datetime fields in ISO format.
 
-    This field validator requires the ISO 8601 date and time format with ``Z``
-    as the time zone (``YYYY-MM-DDTHH:MM:SSZ``). This format is compatible
-    with Kubernetes and the ISO UWS standard and is the same format produced
-    by `safir.datetime.isodatetime`. It should be used when the ambiguous
-    formats supported by Pydantic by default (such as dates and times without
-    time zone information) shouldn't be allowed.
+    This field validator requires the ISO 8601 date and time format, either
+    using ``Z`` as the timezone (``YYYY-MM-DDTHH:MM:SSZ``) or without timezone
```
fajpunk, line 85: `YYYY-MM-DD[THH:MM:SS[.mmm]][Z]` is the clearest description of the acceptable format to me, but that may be because I've been staring at this PR for 10 minutes :smile:

fajpunk (approved):
Nice! Some documentation nits, and a suggestion for one more test

PR #328, opened 2024-11-14
DM-47262: Add Python 3.13 support

rra:
Add Python 3.13 to the supported versions for each package, and add it to the test matrix.


rra:
Abandoning this for now. It will be back later in a different form once dataclasses-avroschema has Python 3.13 support.

PR #334, opened 2024-11-22
DM-47769: Add LinkData model for parsing HTTP Link headers

rra:
Add `safir.models.LinkData`, which parses HTTP `Link` headers for pagination information. This is useful for clients of services that use the database pagination support, including their test suites.


rra:
This little toy model isn't great and doesn't handle the full syntax, but I already had it sitting around and it's enough to use in test suites without having to write a parser. Maybe I should add more disclaimers to the documentation? If we were using requests, we could use its built-in parser, but httpx apparently doesn't have such a thing and there's only one kind-of abandoned Python module that I could find.

fajpunk (approved):
What about calling it `PaginationLinkData`, or something not so verbose but that still indicates it's only useful for pagination headers?  Or, keep it in the `safir.database` namespace to indicate that its only useful there?

rra:
Ah, yes, both of those changes sound good. Done.

PR #340, opened 2024-11-26
DM-47459: Added middleware for converting form post params to lowercase

stvoutsin:
**Summary**
Most IVOA specifications expect service providers to support case-insensitive params, including for `application/x-www-form-urlencoded` POST requests. This PR adds a middleware class `CaseInsensitiveFormMiddleware` which parses the form params and converts them to lowercase.

As part of the changes I've also added a couple tests for the new middleware class, and in the process have broken the `middleware/ivoa_test.py` tests into smaller individual tests. I can revert that if the single test is preferred.

**Tests**
- Added unit tests and tested via nox
- Tested SIAv2 app with new middleware on data-dev

### DIFF
--- a/changelog.d/20241126_175555_steliosvoutsinas_DM_47459.md
+++ b/changelog.d/20241126_175555_steliosvoutsinas_DM_47459.md
@@ -1,0 +1,5 @@
+<!-- Delete the sections that don't apply -->
+
+### New features
+
+- Added CaseInsensitiveFormMiddleware to lowercase handle form post params for VO services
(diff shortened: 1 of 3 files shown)


changelog.d/20241126_175555_steliosvoutsinas_DM_47459.md
```diff
@@ -0,0 +1,5 @@
+<!-- Delete the sections that don't apply -->
+
+### Other changes
+
+- Added CaseInsensitiveFormMiddleware to lowercase handle form post params for VO services
```
rra, line 5: I think this goes under new features.

rra (approved):
This looks great, thank you!

PR #346, opened 2024-12-03
DM-47928: Use case-insensitive form middleware for UWS support

rra:
Convert the UWS support to the newly-added case-insensitive form middleware, which removes the need for annoying workarounds for case-insensitive form parameters. Update the application documentation accordingly.


rra:
This is an initial step towards using the models from vo-models for the rest of the UWS API, and also simplifies the code considerably.

stvoutsin (approved):
Looks great to me

PR #348, opened 2024-12-05
DM-47764: App metrics unit testing helpers

fajpunk:
Add a `MockEventManager` that creates `MockPublishers` which are no-op publishers that record all published payloads. This is returned from the config factory function when `enabled` is `False` and a new config key, `mock`, is `True`.

Payloads are recorded to a list with mixed-in assertion helpers.

### DIFF
--- a/safir/src/safir/metrics/_event_manager.py
+++ b/safir/src/safir/metrics/_event_manager.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import time
 from abc import ABCMeta, abstractmethod
 from datetime import UTC, datetime
-from typing import Generic, TypeAlias, TypeVar, cast
+from typing import Generic, TypeVar, cast
 from uuid import uuid4
 
 import structlog
@@ -27,18 +27,6 @@ from ._testing import PublishedList
 
 P = TypeVar("P", bound=EventPayload)
 """Generic event payload type."""
-
-EnrichedEvent: TypeAlias = P | EventMetadata
-"""Alias to refer to the dynamically created payload+metadata event type.
-
-Since this type is dynamically created, there is no way to name it or make it a
-TypeVar. We could declare it as ``P`` everywhere we need to declare it, but
-that makes it difficult to see when we are using the payload vs the
-payload+metadata types.
-
-This lets us differentiate it from a payload class in the code, and it can be
-narrowed further if we have to access either metadata or payload attributes.
-"""
 
 __all__ = [
     "EventManager",
@@ -66,13 +54,11 @@ class EventPublisher(Generic[P], metaclass=ABCMeta):
         publication.
     """
 
-    def __init__(
-        self, application: str, event_class: type[EnrichedEvent]
-    ) -> None:
+    def __init__(self, application: str, event_class: type[P]) -> None:
         self._application = application
         self._event_class = event_class
 
-    def construct_event(self, payload: P) -> EnrichedEvent:
+    def construct_event(self, payload: P) -> P:
         """Construct the full event as it will be published.
 
         Parameters
@@ -153,7 +139,7 @@ class KafkaEventPublisher(EventPublisher, Generic[P]):
         *,
         application: str,
         manager: KafkaEventManager,
-        event_class: type[EnrichedEvent],
+        event_class: type[P],
         publisher: AsyncAPIDefaultPublisher,
         schema_info: SchemaInfo,
     ) -> None:
@@ -179,7 +165,7 @@ class NoopEventPublisher(EventPublisher, Generic[P]):
     def __init__(
         self,
         application: str,
-        event_class: type[EnrichedEvent],
+        event_class: type[P],
         logger: BoundLogger,
     ) -> None:
         super().__init__(application, event_class)
@@ -204,7 +190,7 @@ class MockEventPublisher(NoopEventPublisher, Generic[P]):
     def __init__(
         self,
         application: str,
-        event_class: type[EnrichedEvent],
+        event_class: type[P],
         logger: BoundLogger,
     ) -> None:
         super().__init__(application, event_class, logger)
@@ -500,7 +486,7 @@ class KafkaEventManager(EventManager):
 
     async def publish(
         self,
-        event: EnrichedEvent,
+        event: P,
         publisher: AsyncAPIDefaultPublisher,
         schema_info: SchemaInfo | None,
     ) -> None:
--- a/safir/src/safir/metrics/_event_manager.py
+++ b/safir/src/safir/metrics/_event_manager.py
@@ -483,7 +483,7 @@ class KafkaEventManager(EventManager):
         schema_info = await self._schema_manager.register_model(model)
 
         # Return the corresponding event publisher.
-        return KafkaEventPublisher[P](
+        return KafkaEventPublisher(
             application=self._application,
             event_class=model,
             publisher=async_publisher,
@@ -600,7 +600,7 @@ class NoopEventManager(EventManager):
         EventPublisher
             An appropriate event publisher implementation instance.
         """
-        return NoopEventPublisher[P](self._application, model, self.logger)
+        return NoopEventPublisher(self._application, model, self.logger)
 
 
 class MockEventManager(EventManager):
--- a/safir/src/safir/metrics/_event_manager.py
+++ b/safir/src/safir/metrics/_event_manager.py
@@ -469,7 +469,7 @@ class KafkaEventManager(EventManager):
         schema_info = await self._schema_manager.register_model(model)
 
         # Return the corresponding event publisher.
-        return KafkaEventPublisher(
+        return KafkaEventPublisher[P](
             application=self._application,
             event_class=model,
             publisher=async_publisher,
@@ -586,7 +586,7 @@ class NoopEventManager(EventManager):
         EventPublisher
             An appropriate event publisher implementation instance.
         """
-        return NoopEventPublisher(self._application, model, self.logger)
+        return NoopEventPublisher[P](self._application, model, self.logger)
 
 
 class MockEventManager(EventManager):
@@ -636,4 +636,4 @@ class MockEventManager(EventManager):
         EventPublisher
             An appropriate event publisher implementation instance.
         """
-        return MockEventPublisher(self._application, model, self.logger)
+        return MockEventPublisher[P](self._application, model, self.logger)
--- a/safir/src/safir/metrics/_event_manager.py
+++ b/safir/src/safir/metrics/_event_manager.py
@@ -103,11 +103,6 @@ class EventPublisher(Generic[P], metaclass=ABCMeta):
             accessed.
         """
 
-    @property
-    def published(self) -> PublishedList[P]:
-        """A list of published event payloads with some test helpers."""
-        raise NotImplementedError("Only available on a MockEventPublisher.")
-
     @staticmethod
     def _ns_to_datetime(ns: int) -> datetime:
         """Convert an `int` number of nanoseconds to a `~datetime.datetime`."""
@@ -606,7 +601,6 @@ class MockEventManager(EventManager):
     This is for use only in unit testing. Don't use it in any deployed
     environment because memory usage will grow unbounded.
 
-
     Parameters
     ----------
     application
@@ -642,4 +636,4 @@ class MockEventManager(EventManager):
         EventPublisher
             An appropriate event publisher implementation instance.
         """
-        return MockEventPublisher[P](self._application, model, self.logger)
+        return MockEventPublisher(self._application, model, self.logger)
--- a/safir/src/safir/metrics/_event_manager.py
+++ b/safir/src/safir/metrics/_event_manager.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import time
 from abc import ABCMeta, abstractmethod
 from datetime import UTC, datetime
-from typing import Generic, TypeVar, cast
+from typing import Generic, TypeAlias, TypeVar, cast
 from uuid import uuid4
 
 import structlog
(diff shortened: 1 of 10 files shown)


safir/src/safir/metrics/_event_manager.py
```diff
@@ -100,6 +103,11 @@ async def publish(self, payload: P) -> EventMetadata:
             accessed.
         """
 
+    @property
+    def published(self) -> PublishedList[P]:
+        """A list of published event payloads with some test helpers."""
+        raise NotImplementedError("Only available on a MockEventPublisher.")
+
```
rra, line 110: I would just omit this definition from the parent class. That moves the error from runtime to be detectable by mypy. Tests can then use an `isinstance` assertion to check that the publisher is of the type they expected and then freely use `.published`. I think that's better than having a property that can fail at runtime after passing a type check.

safir/src/safir/metrics/_event_manager.py
```diff
@@ -176,6 +184,34 @@ async def publish(self, payload: P) -> EventMetadata:
...
+    This is meant to be used in unit tests to enable assertions on published
+    payloads. It should NOT be used in any deployed application instances
+    because memory usage will grow unbounded with each published event.
+    """
+
+    def __init__(
+        self,
+        application: str,
+        event_class: type[AvroBaseModel],
```
rra, line 198:
```suggestion
        event_class: type[P],
```

safir/src/safir/metrics/_event_manager.py
```diff
@@ -556,3 +592,54 @@ async def build_publisher_for_model(
...
+
+    This is used as the implementation of `~safir.metrics.EventManager` when
+    event publication is disabled and mocking is enabled. Like a
+    `~safir.metrics.NoopEventManager`, the event type registrations are still
+    verified to catch errors, but any calls to
+    `~safir.metrics.MockEventPublisher` are recorded for later assertion.
+
+    This is for use only in unit testing. Don't use it in any deployed
+    environment because memory usage will grow unbounded.
+
```
rra, line 608: Nit: stray blank line

safir/src/safir/metrics/_event_manager.py
```diff
@@ -556,3 +592,54 @@ async def build_publisher_for_model(
...
+        model
+            Enriched and configured model representing the event that will be
+            published.
+
+        Returns
+        -------
+        EventPublisher
+            An appropriate event publisher implementation instance.
+        """
+        return MockEventPublisher[P](self._application, model, self.logger)
```
rra, line 645: With the change above to the constructor argument type, I believe you can then omit the `[P]` qualifier here and mypy will figure it out for you.

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+    "NotPublishedConsecutivelyError",
+    "NotPublishedError",
+    "PublishedCountError",
+    "PublishedList",
+    "PublishedTooFewError",
+]
+
+P = TypeVar("P", bound=BaseModel)
+"""Generic event payload type."""
+
```
rra, line 23: Nit: Lots of extra blank lines at the start of this file, I think because of mixing variables and type variables and classes. There's no way to avoid that for `_NotNone`, but I'd lift the other variables and type variables up above the class and `NOT_NONE` definition just to compact things a bit. (We're inconsistent about whether to put globals before or after `__all__`.)

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+    "NOT_NONE",
+    "BaseAssertionError",
+    "NotPublishedConsecutivelyError",
+    "NotPublishedError",
+    "PublishedCountError",
+    "PublishedList",
+    "PublishedTooFewError",
+]
+
+P = TypeVar("P", bound=BaseModel)
```
rra, line 21: I don't think it matters a ton, but shouldn't the bound here be `EventPayload`, just to reduce the chance of errors from passing in some unrelated Pydantic model?

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def __repr__(self) -> str:
+        return "<NOT NONE>"
+
+
+NOT_NONE = _NotNone()
+"""An object to indicate that a value can be anything except None."""
+
+
+ModelDumpList = list[dict[str, Any]]
```
rra, line 43:
```suggestion
ModelDumpList: TypeAlias = list[dict[str, Any]]
```

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+class NotPublishedError(BaseAssertionError):
+    """Some expected items were not published."""
+
+    def __init__(
+        self,
+        expected: ModelDumpList,
+        actual: ModelDumpList,
+        actual_models: list[P],
+        not_found: ModelDumpList,
+    ) -> None:
```
rra, line 104: Nit: Over three parameters, so generally we would force these to be keyword parameters by adding `*,` after `self,`.

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+            f"{pformat(self.not_found)}"
+        )
+
+
+class PublishedList(Generic[P], list[P]):
+    """A list of event payload models with assertion helpers.
+
+    All assertion helpers take lists of dicts as expected items and use the
+    ``model_dump(mode="json")`` serialization of the models for comparison.
+
```
rra, line 121: Nit: stray blank line

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+
+    def assert_published(
+        self,
+        expected: ModelDumpList,
+        *,
+        any_order: bool = False,
+    ) -> None:
+        """Assert that all of the expected payloads were published.
+
```
rra, line 131: Nit: stray blank line

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+    def assert_published_all(
+        self,
+        expected: ModelDumpList,
+        *,
+        any_order: bool = False,
+    ) -> None:
+        """Assert that all of the expected payloads were published, and only
+        the expected payloads, were published.
+
```
rra, line 174: Nit: stray blank line.

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+            Expected events were not published consecutively
+        PublishedCountError
+            A different number of events were published than were expected
+        """
+        if len(expected) != len(self):
+            actual = [model.model_dump(mode="json") for model in self]
+            raise PublishedCountError(expected, actual, self)
+        self.assert_published(expected, any_order=any_order)
+
+    def _unordered(
```
rra, line 201: Maybe `_check_unordered`? Right now the method name is grammatically a noun, which implies that it's returning some piece of data, but it's actually performing an action, so it feels like it should be a verb.

safir/src/safir/metrics/_testing.py
```diff
@@ -0,0 +1,235 @@
...
+        self,
+        expected: ModelDumpList,
+    ) -> None:
+        """Assert that each expected event was published.
+
+        Each published event is matched only once. Taken from the
+        ``unittest.mock`` ``assert_has_calls`` method.
+
```
rra, line 209: Nit: stray blank line. (I wish Ruff would remove these for you.)

rra (approved):
This looks great, thank you! All the comments are minor stuff.

safir/src/safir/metrics/_event_manager.py
fajpunk: `P` is the payload type here, whereas `event_class` is the dynamically smooshed payload and metadata type.

rra: Then the prototype for `EventPublisher` is also wrong, but I'm not sure that's correct. See the prototype for `construct_event`; we're using `P` as the type for the payload.

We don't have a separate type for the smooshed payload and metadata type. It would be ideal if we did, but since it's dynamically constructed, mypy cannot represent that. However, it is a child class of `P`, so using `P` for it is not incorrect, and is more accurate (and more constrained) than saying it's any `AvroBaseModel`.

fajpunk: TLDR: Blech, definitely agreed that its inconsistent. What do you think about using a type alias to refer to the enriched event, like [this](https://github.com/lsst-sqre/safir/pull/348/commits/06ce821c1bc00d5f7673d142e81572e222a2b702)?

I don't love using the `P` parameter to mean both (even though I've already done it :smile: ) because I think it makes the code confusing and obscures the already-tricky difference between the two types. I think that's why I originally (and inconsistently :disappointed: ) declared the enriched type as `AvroBaseModel`, to differentiate it.

As a side note, I don't understand why mypy thinks it's OK that [`construct_event` returns `P`](https://github.com/lsst-sqre/safir/blob/5e1c4f5b6798f99a8a66ed70afea487c885fc609/safir/src/safir/metrics/_event_manager.py#L61-L84) and [`publish` returns the `construct_event` return value but says it's an `EventMetadata`](https://github.com/lsst-sqre/safir/blob/5e1c4f5b6798f99a8a66ed70afea487c885fc609/safir/src/safir/metrics/_event_manager.py#L151-L154), when [`P` is bound to `EventPayload`](https://github.com/lsst-sqre/safir/blob/5e1c4f5b6798f99a8a66ed70afea487c885fc609/safir/src/safir/metrics/_event_manager.py#L28-L28)

...looks like mypy is saying that the return value from `construct_event` is `Any`:
```
async def publish(self, payload: P) -> EventMetadata:
        event = self.construct_event(payload)
        reveal_type(event)
        await self._manager.publish(event, self._publisher, self._schema_info)
        return event
```
```nox > mypy --namespace-packages --explicit-package-bases noxfile.py safir/src safir-arq/src safir-logging/src safir/tests
safir/src/safir/metrics/_event_manager.py:153: note: Revealed type is "Any"
```
It makes sense to me that mypy doesn't know what the type is, but I don't get why mypy  lets us call it something else and then just starts throwing `Any` around.  I'm glad it does in this case, otherwise this wouldn't work :smile: :shrug: 

Fwiw, The `EnrichedEvent` change doesn't help here, it still subs `Any` for the `P`:
```nox > mypy --namespace-packages --explicit-package-bases noxfile.py safir/src safir-arq/src safir-logging/src safir/tests
safir/src/safir/metrics/_event_manager.py:167: note: Revealed type is "Union[Any, safir.metrics._models.EventMetadata]"```

rra: I'm not a fan of the union because I think that will make the typing more tedious, not less. If someone wants to see the enriched fields that we add to the type in the return value from publish, now they'll have to add a cast or an `isinstance` check or mypy will complain that those fields are missing from the `P` branch of the union.

The root problem here is that the semantics you want for the return value from `publish` are intersection, not union. The returned type has to be a child type of *both* `P` and `EventMetadata`, and is therefore a valid instance of both of them. Unfortunately, there isn't a simple way to represent this because `TypeVar` only takes a single bound.

Because of that, I feel like the proposed change is even more inaccurate in the typing, since it semantically says that this model can be either some instance of `EventPayload` *or* `EventMetadata`, but really it has to be both, and the union also loses the strong association between the constructor parameter type and the input type for publish (you could pass a simple instance of `EventMetadata` to `publish` now, which is wrong).

I think the best we can do here is to parameterize `EventPublisher` with two `TypeVar` parameters, one of which has a bound of `EventPayload` and represents the application-side event model and the other of which has a bound of `EventMetadata` and represents the enriched event. Call them `P` and `E`. Then, `__init__` takes a parameter of type `type[E]`, `construct_event` takes `P` and returns `E`, and `publish` takes `P` and returns `E`.
I think that's as good as we can get in representing the type system given the lack of multiple bounds, the lack of a way to define one generic type parameter is a subclass of another, and the fact mypy is always going to bind `E` to `Any` right now anyway.

But the problem with doing *that* is that when you construct the `EventPublisher` class, you have to provide `P` since it cannot be determined from the constructor arguments, which I think means you have to provide `E` as well, and there's no valid value of `E` that you can provide. You end up having to write an explicit `Any`, I think.

This is why I just used `P` for everything when I suggested this type scheme originally. It seems strictly better in terms of type algebra than using `AvroBaseModel`, which is much less constrained. And it is actually *correct*; it's always valid to declare a function returning some class as returning any superclass of that class, and `P` is a superclass of every class involved, so the type system here where `P` is used everywhere is formally correct. It just discards information because it loses the fact that `EventMetadata` has been mixed in and those fields are therefore guaranteed to be present (which is why the return type of `publish`).

rra: I think the critical thing here is to make sure that the resulting type of the publisher object is `EventPublisher[SomeConcreteMetricsEvent]`, not `EventPublisher[Any]`. That does mean that I'm wrong and you can't omit the `[]` when creating the publisher. You do need to ensure that's bound to the input event model (*not* the enriched type).

fajpunk: `P` works for me.

> But the problem with doing that is that when you construct the EventPublisher class, you have to provide P since it cannot be determined from the constructor arguments, which I think means you have to provide E as well, and there's no valid value of E that you can provide. You end up having to write an explicit Any, I think.

Yep, I tried this and this is more or less where I wound up.

> That does mean that I'm wrong and you can't omit the [] when creating the publisher. You do need to ensure that's bound to the input event model (not the enriched type).

Assuming you mean [this](https://github.com/lsst-sqre/safir/pull/348/commits/9efbb16f58ae84300ce664dc22190bd8605c3a89), it doesn't seem to make a difference, `mypy` can figure it out, I think because we cast the `build_publisher_for_model` arg [here](https://github.com/lsst-sqre/safir/blob/tickets%2FDM-47764%2Fmetrics-events-unittest-support/safir/src/safir/metrics/_event_manager.py#L323-L329).

But it doesn't seem to hurt anything to be explicit about it, and I quite obviously have a lot to learn about how `mypy` deals with generics, so I put them back in :)

PR #349, opened 2024-12-10
DM-47986: Convert the UWS library to use the Wobbly backend

rra:
Rather than storing UWS jobs directly in a database, which requires every UWS-based application to manage its own separate database, use the Wobbly service to manage all job storage. This service uses a delegated token to determine the user and service, so considerably less tracking of the user is required.

UWS applications now store the serialized parameter model in the database rather than a list of key/value pairs, and rely on methods on the parameters model to convert to the XML format for the current IVOA UWS protocol.

Add a mock for Wobbly that can be used to test UWS applications without having the Wobbly API available.

Drop the `ErrorCode` enum, since its values were specific to SODA, and instead take the error code as a string. Drop some related exceptions that are not used directly in Safir and are specific to SODA.


rra:
The typing here internally is not great — the parameters type is `Any` throughout much of the internal code for mypy — because I hadn't come up with a good way to do type propagation into the internals. I have subsequently thought of a strategy that should work, but it's a major refactoring, so I'm saving it for a follow-on PR.

rra:
I should also note that I temporarily removed job expiration. I need to think about the right way to handle that between the application and Wobbly and will re-add it in a later PR.

safir/src/safir/uws/_constants.py
```diff
@@ -4,13 +4,11 @@
 
 from datetime import timedelta
 
-from arq.cron import Options
-
 __all__ = [
     "JOB_RESULT_TIMEOUT",
     "JOB_STOP_TIMEOUT",
     "UWS_DATABASE_TIMEOUT",
```
fajpunk, line 10: Should we call this something else now that this isn't really a database worker?

safir/src/safir/uws/_app.py
```diff
@@ -82,13 +71,7 @@ class UWSApplication:
     def __init__(self, config: UWSConfig) -> None:
```
fajpunk, line 71: Really for lines 56-63 (stupid GitHub...): change the docstring for `UWSApplication` to reflect that it calls wobbly rather than works directly with a database.

fajpunk (commented):
This is great, and I think it reenforces that making Wobbly was the right thing to do. Some minor things, mostly documentation stuff.

safir/src/safir/uws/_config.py
```diff
@@ -251,6 +229,16 @@ class UWSAppSettings(BaseSettings):
...
 
+    wobbly_url: HttpUrl = Field(
+        ...,
+        title="Wobbly URL",
+        description="URL to Wobbly UWS job tracking API",
+    )
+
+    database_password: SecretStr | None = Field(
+        None, title="Password for UWS job database"
+    )
```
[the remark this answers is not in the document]
rra: We don't, good catch!

safir/src/safir/uws/_handlers.py
```diff
@@ -249,15 +247,15 @@ async def get_job_error(
...
 ) -> Response:
     job_service = uws_factory.create_job_service()
-    error = await job_service.get_error(user, job_id)
-    if not error:
+    errors = await job_service.get_error(token, job_id)
+    if not errors:
         raise DataMissingError(f"Job {job_id} did not fail")
     templates = uws_factory.create_templates()
-    return templates.error(request, error)
+    return templates.error(request, errors[0])
```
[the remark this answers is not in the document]
rra: I added a comment. Basically, Wobbly supports storing multiple errors, but the VOTable templating code only supports one error and Safir only ever stores one error, so we just assume the first error is the one that matters for now. Eventually we may add proper support for multiple errors if we have a compelling use case.

safir/src/safir/uws/_constants.py
rra: I think the database worker terminology is still mostly accurate, since that arq worker is responsible for storing results in the database and the timeout is for the worker that does that. It now does this via Wobbly, which exposes an HTTPS API to the database, but that's more of a protocol change than a functional change.

PR #357, opened 2024-12-16
DM-48173: Add abstraction for counted paginated queries

rra:
`PaginatedQueryRunner` has a separate `query_count` method to return the count, but since the total entry count isn't included in the `PaginatedList` data structure, it's annoying to pass between components of a service. Either the result plus the count have to be returned as a tuple, or the service has to add its own data structure to wrap `PaginatedList`.

Avoid this by introducing a `CountedPaginatedQueryRunner` and a corresponding `CountedPaginatedList` that always includes a `count` attribute. This can be used by services such as Gafaelfawr that always want to count the total number of entries, either because the table is small or because the count can always be satisfied from the table indices.


rra:
Tested in https://github.com/lsst-sqre/gafaelfawr/pull/1187

fajpunk (approved):
Nice!

PR #365, opened 2025-01-14
DM-48408: Fix documentation of database initialization

rra:
Avoid using the `run_with_asyncio` decorator for the init command-line command, since Alembic doesn't like to run inside an existing event loop. Instead, recommend using an async helper function and `asyncio.run` from the start so that adding Alembic support is simple.


jonathansick (approved):
Nice, thanks for the update.

PR #377, opened 2025-02-11
DM-48838: Add explicit override decorators

rra:
Add the `@override` decorator from typing to methods intended to override parent methods. This makes refactoring safer by assuring there will be a typing error if the parent method changes names.


rra:
This is blocked until we can drop support for Python 3.11.

PR #389, opened 2025-03-12
Add PersistentVolume support to kubernetes mock

athornton:


athornton:
Turned out that for fileserver support to work I had to do it right anyway.  Have another look?

safir/src/safir/testing/kubernetes.py
```diff
@@ -1840,6 +1845,176 @@ async def list_node(
...
+            Ignored, accepted for compatibility with the Kubernetes API.
+
+        Raises
+        ------
+        kubernetes_asyncio.client.ApiException
+            Raised with 409 status if the persistent volume already
+            exists.
+        """
+        self._maybe_error("create_persistent_volume", body)
+        self._update_metadata(body, "v1", "PersistentVolume", "__cluster__")
```
rra, line 1872: I think this is still erroneously setting a `metadata.namespace` value of `__cluster__`.

rra (commented):
Yeah, this looks like the right direction to me. Thank you! Now that there's proper handling for cluster objects, we should move `Namespace` objects over to it as well and remove the special casing for `Namespace` and have it use these same methods.

rra (approved):
Needs a scriv changelog.d file, but other than that, looks good to me. Thanks!

PR #399, opened 2025-05-21
DM-50964: Allow passing Kafka clients to make_manager

rra:
Add an optional `kafka_clients` argument to the `make_manager` method of `safir.metrics.BaseMetricsConfiguration` to simplify the integration of metrics into applications that want to reuse an existing Kafka broker.

The previous recommended approach for applications already using Kafka required manual construction of an `EventManager`, which bypassed all of the configuration-based automatic selection of no-op and mock event managers. Selection of those manager variants is desirable even in an application with an existing Kafka broker, since it may still want to test metrics or disable metrics reporting.

Leave in the manual `KafkaEventManager` construction directions just in case, since this is currently is the only way to provide a pre-existing schema manager. Fix a few bugs in that documentation.


safir/src/safir/metrics/_config.py
```diff
@@ -131,7 +153,11 @@ class DisabledMetricsConfiguration(BaseMetricsConfiguration):
 
     @override
     def make_manager(
-        self, logger: BoundLogger | None = None
+        self,
+        logger: BoundLogger | None = None,
+        *,
+        kafka_broker: KafkaBroker | None = None,
+        kafka_admin_client: AIOKafkaAdminClient | None = None,
```
fajpunk, line 160: You could do something like:
```python
@dataclass
class KafkaClients:
    broker: KafkaBroker
    admin_client: AIOKafkaAdminClient

def make_manager(self, logger: str, kafka_clients: KafkaClients | None):
    etc
```
Then the "both or neither" condition would be verified at the type level... at the expense of a weird interface. I don't have a strong preference either way!

fajpunk (approved):
This is much better, thank you!

safir/src/safir/metrics/_config.py
rra: Oh, yeah, that's better. I was trying to figure out a way to do that and that one didn't occur to me. Thanks!

PR #401, opened 2025-05-21
DM-50964: Support serialization options for Redis storage

rra:
In the `store` method of `PydanticRedisStorage` and `EncryptedPydanticRedisStorage`, support some of the Pydantic serialization options so that applications can reduce the size of serialized objects with large numbers of optional fields. Document that normally these flags should not be used because the default serialization is more future-proof.

This support is desired to reduce the size of query objects stored by the Qserv Kafka bridge with wide returns, since the optional fields for the per-column schema definition otherwise can consume over 100KB of space in the serialized object.


jonathansick (approved):
Good idea, thanks.

PR #402, opened 2025-05-23
DM-50964: Add testcontainers support for Kafka

rra:
Lift the testcontainers containers for Kafka and the Confluent schema registry into a new `safir.testing.containers` module and document them so that they can be used by other packages to test with a real Kafka server.

Extensively refactor and simplify the Kafka test container to remove things we weren't using, update the defaults to match our use cases, and simplify the internal class structure and public API.


safir/src/safir/testing/containers/_kafka.py
```diff
@@ -0,0 +1,341 @@
...
+
+    .. code-block:: python
+
+       from testcontainers.kafka import KafkaContainer
+
+       with KafkaContainer() as kafka:
+           connection = kafka.get_bootstrap_server()
+    """
+
+    _CLUSTER_ID = "MkU3OEVBNTcwNTJENDM2Qk"
```
fajpunk, line 55: I don't _think_ there's anything magical about this value. I know it's used in the Kafka docker docs and a bunch of the random internet posts. We could probably change it if we want to, or say here that its arbitrary.

safir/src/safir/testing/containers/_kafka.py
```diff
@@ -0,0 +1,341 @@
...
+        ----------
+        timeout
+            How long to wait for the container to start before raising an
+            error if Kafka still isn't running.
+        """
+        self._configure()
+
+        # Create a start command that waits for the start script to be created
+        # and then runs it. We have to do it this way because the start script
+        # is not backed into the container and we can't add it to the
```
fajpunk, line 136:
```suggestion
        # is not baked into the container and we can't add it to the
```

fajpunk (approved):

PR #407, opened 2025-05-27
DM-50964: Add parse_log_tuples for structlog testing

rra:
Add a helper function, `parse_log_tuples`, that parses JSON messages logged by structlog, verifies some basic information, and returns the messages in a structured format so that a test suite can verify the logs include the correct information.


rra:
I went back and forth on putting it into `safir.testing.logging` instead. What do you think about that idea? That way we group all the testing code under `safir.testing`. It does mean that applications that only use `safir-logging` would need to depend on `safir` for their test suite, but (a) I'm fine with that I think, and (b) I don't think we have any such applications anyway.

jonathansick:
I think `safir.testing.logging` would be the best story. Is there a way for safir-logging to add to that namespace? Otherwise maybe migrating the module over would be great.

PR #417, opened 2025-07-23
DM-51521: Revise documentation for current templates

rra:
There are some steps users should take after creating a new project via Squarebot, so devote the main tutorial page to that and move the manual process into a separate page with some duplication of steps. Update both for the current templates (uv, Ruff instead of Black, new dependency management method).

Update the README files to point to Phalanx in general and list the Rubin Science Platform and Roundtable as separate cases. Drop the link to the old Roundtable site, since that's no longer useful to users.

Drop the list of Safir features from the README. It's out of date and largely duplicates the documentation, which is already linked from the README.

### DIFF
--- a/docs/user-guide/set-up-from-template.rst
+++ b/docs/user-guide/set-up-from-template.rst
@@ -5,7 +5,7 @@ Creating an app from the template
 #################################
 
 The best way to create a new Safir-based application is with the fastapi_safir_app_ template.
-The quickest way to do this is by running the ``/msg squarebot create project`` Slack command on the Rubin project Slack.
+The quickest way to do this is by running the ``/msg @Squarebot create project`` Slack command on the Rubin project Slack.
 
 For a manual equivalent, see :doc:`template-manual`.
 
@@ -21,7 +21,13 @@ See the `uv documentation <https://docs.astral.sh/uv/getting-started/installatio
 2. Create the project
 =====================
 
-To create a new project, run the command ``/msg squarebot create project`` on the Rubin Observatory internal project Slack.
+In the Rubin Observatory internal project Slack, ask `SQuaRE Bot`_ to create a new project:
+
+.. code-block:: text
+   :caption: rubin-obs.slack.com
+
+   /msg @Squarebot create project
+
 Select the ``FastAPI application (Safir)`` template.
 
 This will ask a series of questions to configure your new repository, and then create it from the fastapi_safir_app_ template.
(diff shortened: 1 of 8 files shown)


docs/user-guide/set-up-from-template.rst
```diff
@@ -4,58 +4,47 @@
...
 
-2. Start the project with the template
-======================================
+2. Create the project
+=====================
 
-Next, you'll actually create the project files using the template.
-
-.. prompt:: bash
+To create a new project, run the command ``/msg squarebot create project`` on the Rubin Observatory internal project Slack.
```
jonathansick, line 24: I'd be tempted to put this in a code-block (if not a prompt but i guess it'd not a traditional command line prompt) just because people will scan for it. Also keep in mind the username formatting:

```rst
.. code-block:: text
   :caption: rubin-obs.slack.com

   /msg @Squarebot create project
```

docs/user-guide/set-up-from-template.rst
```diff
@@ -184,8 +142,8 @@ This development server auto-reloads, so any time you change the code, the serve
 Next steps
 ==========
 
-Now that you have a working application repository, the next steps are to develop your application's logic and interface, and then deploy it to Roundtable.
+Now that you have a working application repository, the next steps are to develop your application's logic and interface, and then deploy it to Phalanx.
 
 To learn learn more about developing Safir-based applications like yours, refer to the :doc:`guides in this documentation <index>` and the `FastAPI documentation <https://fastapi.tiangolo.com/>`__.
 
-To learn how to deploy your application to Roundtable, see the `Roundtable documentation <https://roundtable.lsst.io>`__.
+To learn how to deploy your application to Phalanx, see the `Phalanx documentation <https://phalanx.lsst.io>`__.
```
jonathansick, line 149: Perhaps point directly to the developers section of the Phalanx docs: https://phalanx.lsst.io/developers/index.html

jonathansick (approved):
Thanks for doing this update. We should also touch up https://phalanx.lsst.io/developers/create-an-application.html because I forgot to update that when I rushed to migrate the bots last fall.

I wondered if we could combine the two pages using [tabs](https://sphinx-design.readthedocs.io/en/latest/tabs.html) to fork certain steps between the bot or local setup flows, but I'm not sure it'd work.

rra:
Thanks! I will get Phalanx as well.

Using tabs for the two guides is a great idea, but since the manual process has several more steps, I'm not entirely sure how to set it up in a way that feels visually balanced. I may be over-thinking it, but I'm going to bail for now. Good thought, though, and maybe I'll tackle it the next time I revisit this.

PR #420, opened 2025-07-28
DM-51984: Use stop instead of close for FastStream

rra:
Close FastStream brokers with `stop` instead of `close` and require FastStream 0.5.44 or later to avoid deprecation warnings.


fajpunk (approved):
Thank you!

PR #433, opened 2025-08-21
DM-52238: Add lower bounds on Google modules

rra:
Updating dependencies on vo-cutouts tried to downgrade the Google API modules to an ancient version because they pin an older major version of cachetools and the latest version of tox depends on the new version. Ensure this cannot happen by setting a version floor for the Google API modules.


rra:
On further thought, this doesn't really need review, since it's a simple dependency change. I'm going to go ahead and merge it.

PR #438, opened 2025-08-28
Accept regular RFC-3339 strings for parse_isodatetime()

athornton:


rra (changes requested):
It's a good point that this function is probably misnamed since the change in Safir 8.0.0, but we do need the function to do exactly what it does. That's why there's an explicit test that this change fails.

PR #445, opened 2025-09-08
DM-52438: Convert to uv workspaces

rra:
Move the `safir` subdirectory to the top level and use uv workspaces to manage safir-arq and safir-logging. Commit the top-level `uv.lock` file following best-practices recommendations from the uv developers for library packages. This freezes the dependencies for development but has no effect on the dependencies of the uploaded library packages.

Adopt nox-uv and use its capabilities alongside workspaces to simplify `noxfile.py`. nox-uv handles installing extras and dependency groups directly, so there is no longer a need to handle that manually in the `noxfile.py`.

For now, let the uv version float in GitHub Actions. This runs a small risk that new uv releases will break tests in ways that are unrelated to a given change, but we've had good luck with new uv versions so far. If this becomes a problem in practice, we can adopt a more complicated dependency management system with a script to update this version, similar to what we use for service packages.


jonathansick (approved):
Looks all properly set up and workspaces make many of the development and configurations tidier. We discussed in stand-up that the goal of this work isn't related to making Safir particularly easier to install into different client environments, so no worries about that.

One thing that came up from my reading is that I'm not sure the "root+packages" layout is absolutely required, although that's the one uv documents. [Here's one that consists solely of "packages" from their test suite.](https://github.com/astral-sh/uv/tree/main/scripts/workspaces/albatross-virtual-workspace) I agree that treating "safir" as the root makes much sense here. On the other hand, it'd save moving modules around and might make things easier to reason if we break up Safir more in the future.

rra:
Oh, good catch on the example with all workspace members in subdirectories.

For now, until we do the rethink of Safir, I think I'd rather keep the main library at the top level because it saves some typing as we work on it for now and it fits the current conceptual model (or at least mine) of a Safir library that has a few components split off for technical reasons but where most users don't need to know about that implementation detail.

I suspect that the result of the general rethink will be a fair bit of moving things about anyway, and I'm happy to unwind the current layout once that happens, if a different layout makes more sense.

PR #448, opened 2025-09-12
DM-52391: Move Sentry exception functionality into `SlackException`

fajpunk:
I went back and forth on this, but I ended up liking it better than a totally separate `SentryException`, or a new `UniversalException` or something like that.

Pros:
* Apps will get sentry-enhanced Safir exceptions without having to modify their code

Cons:
* The Sentry stuff pollutes the `slack` namespace with stuff that's not directly related to Slack.

With the way this is structured, we should be able to split out the `sentry_sdk` dependency into an extra if we want.


src/safir/slack/sentry.py
```diff
@@ -0,0 +1,20 @@
...
+    """A collection of metadata to add to Sentry events.
+
+    The overrideable `~SlackException.to_sentry` method returns an object of
+    this type.
+    """
+
+    tags: dict[str, str]
+    contexts: dict[str, dict[str, Any]]
+    attachments: dict[str, Any]
+    username: str | None
```
rra, line 20: Needs docstrings for the members (great spot to put a short version of the hints from the documentation for when to use a tag vs. a context vs. an attachment). Also, per above, default values for tags, contexts, and attachments, and maybe of `None` for username as well, seem appropriate.

rra (approved):
Just a few minor things, but overall this looks great.

PR #451, opened 2025-09-22
Fix initialize_sentry typing

fajpunk:
The type in `**kwargs: <type>` is the type of one of the values, not of the kwargs dict :facepalm: Someday I will retain this fact.

In the meantime, I don't think there is any way to keep the type info of the params to `sentry_sdk.init`. There is a way to keep type info from a wrapped function's params, and info from a wrapped function's params plus additional  positional args, but no way to keep info from a wrapped function's params _minus_ some kw params:

https://peps.python.org/pep-0612/#concatenating-keyword-parameters


rra (approved):
Yeah, I don't know of a good way around the `Any` here. I think it might be possible with a sufficiently creative use of a wrapper function *if* one is okay with exposing the kwargs that we override to the typing, which probably would be fine, but it's complicated and ugly and I'm okay with living with an `Any` here for the time being.

PR #556, opened 2026-07-06
DM-55366: Support pessimistic database connection handling

### LINKED ISSUE #552: Support pool_pre_ping and pool_recycle in create_database_engine and db_session_dependency
## Problem

Safir apps using `create_database_engine` / `db_session_dependency` cannot enable SQLAlchemy's `pool_pre_ping` or set `pool_recycle`, so they are exposed to stale-connection errors when a pooled connection is closed underneath the pool (idle connection dropped by a network intermediary, connection pooler, or a DB restart/failover).

Ook hit this on a normal request — the pooled connection was already closed when the per-request transaction tried to begin:

```
File ".../asyncpg/connection.py", line 1605, in _check_open
    raise exceptions.InterfaceError('connection is closed')
asyncpg.exceptions._base.InterfaceError: connection is closed

The above exception was the direct cause of the following exception:
  ...
File ".../sqlalchemy/dialects/postgresql/asyncpg.py", line 520, in _prepare_and_execute
    await adapt_connection._start_transaction()
  ...
sqlalchemy.exc.InterfaceError: (sqlalchemy.dialects.postgresql.asyncpg.InterfaceError)
    <class 'asyncpg.exceptions._base.InterfaceError'>: connection is closed
```

The connection object itself reported it was closed (`_check_open` failed before any I/O), which is the classic signature of a dead connection served from the pool rather than a query/timeout/auth failure or an unreachable database.

## Why this can't be addressed downstream today

`create_database_engine` (`src/safir/database/_connection.py`) forwards only `connect_args`, `isolation_level`, `max_overflow`, `pool_size`, `pool_timeout` to `create_async_engine`. `DatabaseSessionDependency.initialize` (`src/safir/dependencies/db_session.py`) mirrors that same set. Neither `pool_pre_ping` nor `pool_recycle` is reachable, and they cannot be set via `connect_args` (those go to the asyncpg driver, not the pool). So a Safir app cannot apply the standard SQLAlchemy remedy without abandoning these helpers and building its own engine/session.

## Proposed solution

Add `pool_pre_ping: bool | None = None` and `pool_recycle: int | None = None` parameters to:

1. `create_database_engine(...)` — forward to `create_async_engine` when set (omit when `None` to preserve current defaults).
2. `DatabaseSessionDependency.initialize(...)` — accept the same two parameters and pass them through to `create_database_engine`.

`pool_pre_ping=True` issues a lightweight liveness check on checkout and transparently replaces dead connections, eliminating this error. `pool_recycle` lets apps proactively recycle connections below a known proxy/server idle timeout.

Optional follow-up for discussion: whether `pool_pre_ping=True` should become the Safir default, since the cost is one cheap round-trip per checkout and the benefit is resilience to idle-connection drops that nearly every deployment environment can cause.

## Alternatives considered

- **Local workaround in the downstream app** — bypass the Safir helpers and call `create_async_engine(..., pool_pre_ping=True)` directly with a custom session dependency. Works, but every Safir app would re-implement it and lose the shared convention.
- **Do nothing** — the error is transient/self-healing, but it surfaces as periodic 500s after idle periods or DB restarts.

## References

- SQLAlchemy "dealing with disconnects" / pessimistic disconnect handling: https://docs.sqlalchemy.org/en/20/core/pooling.html#disconnect-handling-pessimistic
- Error code reference (`rvf5`): https://sqlalche.me/e/20/rvf5

rra:
Add the `pool_pre_ping` and `pool_recycle` parameters to `create_database_engine` and the database session FastAPI dependency with the same meanings and defaults as in SQLAlchemy. Recommend using `pool_pre_ping=True` in applications that only use the database intermittently.

Add `InterfaceError` to the exceptions caught and retried by `@retry_async_transaction` so that it can be used to protect against database connection drops.

Fixes #552

### DIFF
--- a/changelog.d/20260706_143253_rra_DM_55366.md
+++ b/changelog.d/20260706_143253_rra_DM_55366.md
@@ -1,0 +1,4 @@
+### New features
+
+- Add `pool_pre_ping` and `pool_recycle` to `create_database_engine` and `db_session_dependency` with the same meanings as in SQLAlchemy. Recommend setting `pool_pre_ping=True` in applications that only use the database intermittently.
+- Add `InterfaceError` to the exceptions caught and retried by `@retry_async_transaction` so that this decorator can also be used to protect against database connection drops.
--- a/docs/user-guide/database/dependency.rst
+++ b/docs/user-guide/database/dependency.rst
@@ -1,3 +1,5 @@
+.. py:currentmodule:: safir.dependencies.db_session
+
 ############################################
 Using a database session in request handlers
 ############################################
@@ -33,14 +35,15 @@ You must also close the dependency during application shutdown.
 
    app = FastAPI(lifespan=lifespan)
 
-`~safir.dependencies.db_session.DatabaseSessionDependency.initialize` takes the same optional parameters as `~safir.database.create_database_engine`.
+`DatabaseSessionDependency.initialize` takes the same optional parameters as `~safir.database.create_database_engine`.
+See :ref:`db-engine-parameters` for more details.
 
 As with some of the examples above, this assumes the application has a ``config`` object with the application settings, including the database URL and password.
 
 Using the dependency
 ====================
 
-Any handler that needs a database session can depend on the `~safir.dependencies.db_session.db_session_dependency`:
+Any handler that needs a database session can depend on `db_session_dependency`:
 
 .. code-block:: python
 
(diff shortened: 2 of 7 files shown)


jonathansick (approved):
Excellent, thanks for handling these changes.

